Issue #39, opened 2014-02-24
Add ldap authorization


HighwayStar:
It will be very usefull feature for private installation in organisation that already has ldap user base. There is node-ldapauth project on github that can be used to implement it.

riclmb:
Hello,
how do I do to obtain a deposit ShareLaTeX also including and directly LDAP feature? how I should organize myself with Git repositories? command git remove ang after git pull ?
thank you in advance. Best Regards

phuicy:
@riclmb 
It worth noticing the LDAP option at the bottom of
https://github.com/sharelatex/web-sharelatex/blob/master/config/settings.defaults.coffee

henryoswald:
closing because it is now available as part of ShareLaTeX server pro

ralgozino:
Any chance of getting LDAP in the on premise version?

henryoswald:
We are going to be keeping it as part of the Pro package, there is only so much we can realistically make free.

ralgozino:
Ok! it is totally understandable and makes sense. Thanks for the fast answer!
Cheers!

kemnitzs:
If you are looking for an example implementation that uses ldapjs in the server component have a look at.

https://github.com/worksasintended/overleaf_ldap

PR #103, opened 2014-08-27
Ldap Support for Sharelatex

heukirne:
First code for how to solve #39 problem

I put the ldap user as password in mongo to ensure email and ldap user pair.
I can't put the ldap password in mongo because it could change anytime.
The problem here is that one ldap user could have multiple email accounts on sharelatex (not a big issue, but could be avoid)

Any ideia on how to improve it?


riclmb:
Hello,
I'm very interest to deploy authentication ldap on my installation; but I do modifiy and copy the nine files that you have changed; is that what I do ?
because, actually I installed sharelatex and it's work ! after I do copy directory app/ of fork web-sharelatex ....how I do it with the fork sharelatex ?
Thanks.

riclmb:
In other words,  because it is not very clear to me,
how do I do to obtain a deposit ShareLaTeX also including and directly LDAP feature? how I should organize myself with Git repositories? command git remove ang after git pull ?
thank you in advance. Best Regards

heukirne:
Hi @riclmb ,
Just use this git repo https://github.com/heukirne/web-sharelatex for LDAP auth
Only change the config file and its done.

gentakojima:
This works with 0.1.3. We had to add a redirect from login to register form in Nginx, because Sharelatex redirects to Login if you access a project URL.

Whis will work with 0.1.4?

heukirne:
This redirect is a workaround to avoid register and login for LDAP user. I don't understand for what this "0.1.4" refer to.

gentakojima:
Thank you for replying so fast.

Yes, you are redirecting login to register, but I mean that sometimes you end up in login anyway, because there must be somewhere else where there is a redirection to login that you are not managing well. We had to manually add a redirect to Nginx to make sure that login is never reached.

By 0.1.4 I mean the current release version of Sharelatex, as stated on https://github.com/sharelatex/web-sharelatex/releases

We installed the version 0.1.3 back then, and then used your repo with the instuctions from @littlejo comment. LDAP works now, but now we are afraid to update to 0.1.4. Any recommendations?

heukirne:
Stay in this version, soon there will be a official ldap support in sharelatex as explained in https://github.com/sharelatex/web-sharelatex/pull/103#issuecomment-53861402

heukirne:
LDAP now is a ShareLaTeX Server Pro feature
https://www.sharelatex.com/university/onsite.html

PR #134, opened 2015-01-14
Pdfng reload on error

briangough:
Catch errors from pdfjs where possible, and use a 60 second timeout to kill pdfjs and force reload of document if page does not load or render.

If document has to be reloaded more than twice, show the "Compile failed message" so user has to click recompile manually.

### DIFF
--- a/public/coffee/ide/pdfng/directives/pdfRenderer.coffee
+++ b/public/coffee/ide/pdfng/directives/pdfRenderer.coffee
@@ -54,17 +54,19 @@ define [
 				@document.then (pdfDocument) ->
 					pdfDocument.getDestinations()
 
-# Not available in pdf.js-1.0.712, in later versions there is a direct
-# call for this - we should use it as soon as it is available in a
-# stable version
 			getDestination: (dest) ->
+				# There is a direct method for this in pdf.js but it is not
+				# available in pdf.js-1.0.712. Use the following workaround of
+				# getting all the destinations and returning only the one we
+				# want.
 				@destinations = @document.then (pdfDocument) ->
 					pdfDocument.getDestinations()
 				return @destinations.then (all) ->
 					all[dest]
 				, (error) ->
 					console.log 'ERROR', error
-
+				# When we upgrade we can switch to using the following direct
+				# code.
 				# @document.then (pdfDocument) ->
 				# 	pdfDocument.getDestination(dest)
 				# , (error) ->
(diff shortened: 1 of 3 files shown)


public/coffee/ide/pdfng/directives/pdfRenderer.coffee
```diff
@@ -57,15 +62,20 @@ define [
...
+				, (error) ->
+					console.log 'ERROR', error
 
-				@document.then (pdfDocument) ->
-					pdfDocument.getDestination(dest)
-
+				# @document.then (pdfDocument) ->
+				# 	pdfDocument.getDestination(dest)
+				# , (error) ->
+				# 	console.log 'ERROR', error
```
jpallen: Is this code dead now? If so we should remove it fully I think.

public/coffee/ide/pdfng/directives/pdfRenderer.coffee
```diff
@@ -118,17 +129,38 @@ define [
...
+				timedOut = false
+				timer = $timeout () =>
+					Raven.captureMessage?('pdfng page load timed out after ' + @PAGE_LOAD_TIMEOUT + 'ms')
+					# console.log 'page load timed out', pagenum
+					timedOut = true
+					@spinner.stop(element.canvas)
+					# @jobs = @jobs - 1
+					# @triggerRenderQueue(0)
+					this.errorCallback?('timeout')
+				, @PAGE_LOAD_TIMEOUT
```
jpallen: Do we need to do anything to stop the render that is potentially still going on in the background? If not, why not?

jpallen:
No obvious issues from just reading the code, but I don't fully understand the context of some of the changes. If it's tested on your end, I'd say it's good to go.

public/coffee/ide/pdfng/directives/pdfRenderer.coffee
briangough: On successful page load the timeout is immediately cancelled before any rendering is started so this should be ok as only one page is rendered at any time.

public/coffee/ide/pdfng/directives/pdfRenderer.coffee
briangough: The commented code will be what we use when we upgrade to a newer version of pdfjs with a direct method for getDestination, I'll change the comment to make that clearer.

PR #137, opened 2015-01-16
Fix pdfng progress bar

briangough:
Progress bar now tracks the download of the whole document.

pdfDocument.getDownloadInfo() gives a promise which resolves when the document is fully loaded - so we use this to emit a 'loaded' event which hides the progress bar and prevents it appearing to go backwards.


jpallen:
Looks ok in code. Do you know why it originally went backwards? I would expect progress to be reset to zero when a new PDF is loaded, but I don't know why this would cause it to oscillate up and down like it did.

briangough:
I think it went something like this:

Because the progress bar never got hidden/removed, when a reload started the CSS transition took it slowly down to zero. But while that was happening the actual position was going up, so then it started to go up before it got to zero.    

Also the finished loading event came after the first byte range request was completed, not the top, so it never go to the top and disappeared part way through.

PR #138, opened 2015-01-16
use font-face for pdfjs rendering

briangough:
For compatibility with pdfListView  I think we should switch back to using the browser font renderer to display the pdf.  It is a slower on rendering for figures, but does give nicer looking anti-aliased text.


jpallen:
If this the behaviour at the moment anyway in the old viewer? Your call on this, I think you understand the implications better than me. So long as it doesn't cause any hangs or overly noticeable slow down this seems like a good call.

briangough:
Let's test it in your browser this afternoon. 

I checked and the existing pdfListView definitely uses this setting (the default in pdfjs) that I propose to revert to.  So  on documents with a lot of fonts it should not be substantially different than it is with pdfListView.

Issue #124, opened 2014-12-10
Scroll on large images


metbosch:
When you upload a large vertical image the right scroll don't appear. This issue causes that download button doesn't appear. Moreover, only the upper image part can be seen.
![captura de pantalla 2014-12-10 15 08 29](https://cloud.githubusercontent.com/assets/4162025/5377120/89a37fe6-807e-11e4-9805-c19ddb1781a9.png)

jpallen:
Fixed in #140

PR #140, opened 2015-01-19
Allowed scroll on binary-file container

metbosch:
Trying to fix issue #124.
I set overflow in auto mode to allow large images. It seems to be the simplest way and the expected site comportment.


jpallen:
Thank you! Sorry this got neglected for a while.

PR #145, opened 2015-01-26
Fix performance

briangough:
Avoid calling scope.$apply in the following cases

 - editor highlights (on mousemove in editor)
 - closing context menu (on scroll in editor)
 - background saved file check (every second)

Also fixes attrs.layoutDisabled watch on undefined.

### DIFF
--- a/public/coffee/ide/directives/layout.coffee
+++ b/public/coffee/ide/directives/layout.coffee
@@ -103,12 +103,12 @@ define [
 
 					resetOpenStates()
 					onInternalResize()
-					
-					scope.$watch attrs.layoutDisabled, (value) ->
-						if value
-							element.layout().hide("east")
-						else
-							element.layout().show("east")
-							
+
+					if attrs.layoutDisabled?
+						scope.$watch attrs.layoutDisabled, (value) ->
+							if value
+								element.layout().hide("east")
+							else
+								element.layout().show("east")
 		}
 	]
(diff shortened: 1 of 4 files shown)


public/coffee/ide/editor/controllers/SavingNotificationController.coffee
```diff
@@ -13,21 +13,25 @@ define [
...
+						newStatus[doc_id] = oldStatus[doc_id]
+						newStatus[doc_id].unsavedSeconds += 1
 					else
-						$scope.docSavingStatus[doc_id] = {
+						newStatus[doc_id] = {
 							unsavedSeconds: 0
 							doc: ide.fileTreeManager.findEntityById(doc_id)
 						}
 
+			if _.size(newStatus) or _.size(oldStatus)
```
jpallen: I don't understand what this is testing. Can you make it explicit with a comment or refactor?

public/coffee/ide/directives/layout.coffee
```diff
@@ -103,12 +103,12 @@ define [
...
 					onInternalResize()
-					
-					scope.$watch attrs.layoutDisabled, (value) ->
-						if value
-							element.layout().hide("east")
-						else
-							element.layout().show("east")
-							
+
+					if attrs.layoutDisabled?
```
jpallen: There is no chance this is updated dynamically at some point, but we'll miss it because we didn't watch it at the beginning?

PR #146, opened 2015-01-26
Fix pdfng no background digests

briangough:
Add missing scope.apply's needed to fix up the new pdf viewer when there are no digests running in the background.  

Now queues redraws/relayouts so they only occur once.


jpallen:
Tested the disconnect/saving handler and looks good.

Issue #148, opened 2015-01-29
Hijack account for private editing


317070:
Say user A has a document which he wants to edit privately with user B.
User B has no sharelatex account yet, but user A can add him with the share interface. He only needs to add his mail address.
User A then mails user B to tell him about the repository and that they can co-operate online.
Attacker C creates an account with B's mail address before B does, and can access the private document. There is no verification whether user C actually controls the mail address.

It might be beneficial to enforce mail verification before allowing users to access privately shared documents? (See also #136)

jpallen:
I think the solution to this is to share projects with email addresses with a unique token in the URL. I.e. if I share a project with user B, we create a holding account for them, with a unique token, and then email out the link to the project with that unique token attached. So regardless of what email they eventually sign up with, they get access to that project, and we guarantee that only the recipient of the original share email can access the project.

How does that sound? Thoughts and comments welcome. This is an important part of the UX and security of ShareLaTeX, so we should make sure we get it right!

317070:
It does sound like a good solution.

But I wonder if it would be possible to invite people to work on a common project besides mailing them. I for one like the slack we use at our lab.

Wouldn't it be possible to create a one-time-only-invite-url for instance?

jpallen:
Yes, one-time-only-invite-url was essentially what I had in mind. If you invite someone by email we'll send that URL to their email address, but I suppose there's no reason we couldn't give it to you to share manually as well.

henryoswald:
we have a one time token handler which was recycled from the password reset process that would work well for this - https://github.com/sharelatex/web-sharelatex/commit/fe30571893e512916d702ea246372b195050ca1a

Issue #136, opened 2015-01-15
Email confirmation


msimonin:
Hi, 

Does Sharelatex support email confirmation  : the user need to confirm  his email adress (for instance using a confirmation link) to finish the registration process.

henryoswald:
It used to but does not any more. Hopefully we will not need it in the future but having the option would be good.

msimonin:
Actually I'm thinking about adding the option. 

Here is my initial idea : 
* In the method ```registerNewUser``` of ```UserRegistrationHandler.coffee``` we change the email sent to add a confirmation link embedding a confirmation token. We flag the user as unconfirmed in the db and store to confirmation token. If email validation option isn't enabled, we just flag the user as confirmed.
* Add a route ```/confirm?email=xxx&token=xxx```
which validates the user (put the confirm flag to true)
* For all operations that requires the user to be authenticated we check the confirm flag as well. We patch the ```SecurityManager.requiresLogin``` method this purpose.

msimonin:
Since it has been referenced in #148 I'll propose a PR for that.

PR #149, opened 2015-01-30
Email Verification

msimonin:
This PR is a first step to email verification. The missing feature has been referenced in #136 and #148.
In my use case, users are split into two  categories : internal and external. 
internal can create/manage projects, external can't create/copy project. This behaviour is handled by the first commit. ```security.authorizedDomains``` in the config file is an array of email domains corresponding to internal users.
Moreover in my use case I want email verification (1) at creation time (2) at modification time. This is adressed by the second commit. The email verification can be disabled in the settings (```security.emailVerification```)

Those commits are subject to changes, I only want to share this piece of code a start a discussion here about the feature.

### DIFF
--- a/app/coffee/Features/ConfirmEmail/ConfirmEmailController.coffee
+++ b/app/coffee/Features/ConfirmEmail/ConfirmEmailController.coffee
@@ -1,0 +1,46 @@
+ConfirmEmailHandler = require("./ConfirmEmailHandler")
+ConfirmEmailTokenHandler = require("./ConfirmEmailTokenHandler")
+RateLimiter = require("../../infrastructure/RateLimiter")
+UserController = require("../User/UserController")
+UserUpdater = require("../User/UserUpdater")
+UserGetter = require("../User/UserGetter")
+logger = require "logger-sharelatex"
+
+module.exports =
+
+	renderCreatePasswordForm: (req, res)->
+		res.render "user/createPassword", 
+			title:"create_password"
+			emailVerificationToken:req.query.emailVerificationToken
+
+	renderUpdateEmail: (req, res)->
+		ConfirmEmailTokenHandler.getEmailFromTokenAndExpire req.query.emailVerificationToken, (err, emails)->
+			if err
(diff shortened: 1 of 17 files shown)


jpallen:
Thanks for this pull request, there's clearly a lot of work gone into it! Apologies for taking a while to response

I think there are a few things that this PR addresses, and they might be worth talking about in different issues:

1) Defining custom permissions schemas (as you have done). This is always going to be tricky to give people the flexibility they need.
2) Checking that an email owns an email address when we perform actions based on that assumption. We can certainly do this with email verification, but it might be overkill from the user's perspective when something like an anonymous token send out to that email address when sharing a project would establish that they have access to it.
3) Full blown email verification. We had this previously at ShareLaTeX, but my experience is that this should be avoided if at all possible since it creates a clunky and slow workflow for the user.

@henryoswald and I have discussed these things quite a bit internally and our feeling is that if we try to manage this all with lots of if statements 'i.e. if email registration turned on', the code will get very messy and out of control. Instead, we'd like to work on a modular system for authentication and account management so that people can swap out one scheme for another, but each scheme has simple code. In this case you might write your own modified scheme that handles the `authorizedDomains` logic for internal vs external users. How does this sound from your side?

msimonin:
Hi @jpallen, 

I fully agree your point. 
From our side, we are evaluating sharelatex at [Inria](http://www.inria.fr/en/) so we started to make this ad hoc development. It is mainly due to internal security policy we have.

Keeping the code clean and maintainable is for sure mandatory. I like the idea of a modular authentication and account management system, but it's not very clear to me how the design will look like. Have you think about it in more detail ?

jpallen:
@msimonin We're currently building out an account system using LDAP for authentication, which will be the first proof of concept of the modular authentication system. We'll try to keep everyone who is interested in the loop, but it will be a little while before we have anything to release still.

I'd be happy to have a chat about what you need from ShareLaTeX, and have a more detailed discussion about things if you're interested. Drop me an email at james@sharelatex.com if so.

Issue #133, opened 2015-01-12
Adblock Plus Extension Prevents Users from Logging In and Registering


manoyes:
**Edit**: It actually it doesn't appear to affect sharelatex.com...the version of Sharelatex on our servers is a several months old (@82d2ba67890 for web-sharelatex), perhaps this isn't an issue anymore?

First: I can't post grunt logs because it's from a production server with work-related documents. However, this is highly repeatable and affects both the open source version ~~and sharelatex.com~~.

Using the "Adblock Plus" extension in Chrome and Firefox with the EasyPrivacy list enabled makes the "username" and "password" fields disappear on the "Log In" and "Register" pages:

![sharelatex-login-error](https://cloud.githubusercontent.com/assets/1971456/5705461/3384bd70-9a3f-11e4-9a49-611b4a3f7881.png)

In the blockable items list (Select Adblock Plus icon in Firefox > Open Blockable Items), the single blocked URL is:

**Address**: http:SERVER_URL/js/main/event-tracking.js?fingerprint=ALPHANUMERIC_STRING
**Type**: script (blocked)
**Document Source**: SERVER_URL (first party)
**Filter in effect**: -tracking.js?
**Filter Source**: EasyPrivacy

I'm fairly certain it's a naming issue - [EasyPrivacy](https://easylist-downloads.adblockplus.org/easyprivacy.txt) probably blocks any script with "-tracking.js?" in its name. 

A workaround is simply to disable Adblock Plus on your Sharelatex site, but it wasn't immediately obvious to me that Adblock Plus was the issue, as might be the case for a new user who abandons first-time registration.

A fix may be to just change the name of event-tracking.js.

henryoswald:
Thanks for the report. Would it be possible for you to try out a few different names for event tracking to see what works. The name will need to be changed in public/coffee/ide.coffee and public/coffee/main.coffee

jpallen:
Hopefully closed by #159.

PR #159, opened 2015-03-17
Fix for adblock blocking loginscreen.

SibrenVasse:
Rename file to avoid block by adblocker. See issue #133. Works for me.


jpallen:
Good catch and fix, thanks!

Issue #58, opened 2014-03-20
Spell check splits words half way through


jpallen:
For example, the Slovenian word 'povprečje' is split into two words by the č character. At the moment I think we whitelist word characters (i.e. a-zA-Z) and use a special Regex code to match all 'alphanumeric' characters. This works for some accents, but clearly not all. Maybe we should switch to a black list, and only split on characters which are not _, {, }, \, etc.

kruczjak:
Same thing with Polish words, splitted by Polish characters.
For example: 
![selection_028](https://cloud.githubusercontent.com/assets/3147506/6657103/3a8b77ca-cb41-11e4-9390-ec138baa4af6.png)

jpallen:
We need an updated regex that matches these characters as 'word' characters. I failed to find a good solution, but suggestions are welcome from anyone with more experience of this.

henryoswald:
should be fixed now

PR #162, opened 2015-03-23
Update word regex for spellcheck

mickaobrien:
Adds [Latin Extended-A](https://en.wikipedia.org/wiki/Latin_Extended-A) characters to word regex to prevent splitting words half way through on characters like ąęćółżźńś.

Should close issue #58.


jpallen:
Brilliant, thanks!

PR #166, opened 2015-04-15
Suggested API endpoint for JSON-response with projects and tags

jboysen:
Suggested API endpoints for receiving projects and tags. This pattern can form basis for other endpoints as well.

These endpoints uses the BasicAuth pattern for authentication, see app/coffee/Features/Authentication/AuthenticationManager.coffee.

We're of course willing to change the suggested solution if requested.

### DIFF
--- a/app/coffee/Features/Authentication/AuthenticationManager.coffee
+++ b/app/coffee/Features/Authentication/AuthenticationManager.coffee
@@ -3,6 +3,7 @@ User = require("../../models/User").User
 {db, ObjectId} = require("../../infrastructure/mongojs")
 crypto = require 'crypto'
 bcrypt = require 'bcrypt'
+logger = require("logger-sharelatex")
 
 module.exports = AuthenticationManager =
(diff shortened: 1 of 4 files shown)


jpallen:
This is a really good start to a conversation about an API, thanks. My first comment is that any API we have will almost certainly not use direct username, password combinations as the authentication mechanism. Something like OAuth is much more secure.

It's not clear from the code what sort of schema you're expected to be returned from the end points. Is it possible to quickly document (via an example or two), the return format of the data. Then we can discuss if this is appropriate and make sure it will be consistent with any future end points (I know that we have a lot of inconsistencies in naming/camelCase/under_scores in our internal APIs at the moment but we want to make sure a public API is consistent).

I've create an API 'mega thread' at sharelatex/sharelatex#299 where it might be better to direct any further discussion about this so that others can see it and join in.

henryoswald:
It would also be good to know what people are planning on doing with the API, knowing the use cases should help us build something useful.

PR #169, opened 2015-04-21
Un-fade editor when changing theme

bauerj:
Fixes sharelatex/sharelatex#232

This hides the fade for a second and then shows it again. That should be enough to see if you like the color scheme.


bauerj:
@henryoswald ping

bauerj:
:disappointed:

henryoswald:
Sorry for being slow. This method of doing the fading isn't going to work sorry as it uses jquery to modify an angular state. As it stands it breaks the auto hide menu. Sorry!

bauerj:
Okay, no problem.

PR #176, opened 2015-05-03
Public registration

jonathanverner:
# Summary
Implement a public registration page, optionally requiring confirmation & restriction of logins to emails in a particular domain.

# Detailed description
If the  `allowPublicRegistration` variable is set to true (in settings.coffee), the original registration form is enabled allowing anyone to register for an account. 

Moreover, if the `requireRegistrationConfirmation` variable is set to true, new registrations are not activated by default (the activated state is saved in the user attribute `confirmed` in the mongo dbase) but people have to click on a link that is sent to their e-mail address (or enter the confirmation token into a form found at /confirm). This proves that they really control the e-mail address they logged in with.  

The way confirmation is implemented is by reusing the `authToken` user attribute: when a new user is created, a new random token is generated and embedded into the link. When the user confirms the registration by following the link, this token is cleared.

If the `signupDomain` variable is set then registration of new users is restricted to emails in the
given domain.

# Caveat

First of all I'd like to thank you (sharelatex authors) for this excellent piece of software! Also I am new to the code base and I am not sure what the conventions are and whether the way I implemented the features is the right way. Also I am not sure whether the features are something you are interested in at all. So feel free to reject this :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,9 +64,6 @@ public/minjs/
 public/js/main.js
 Gemfile.lock
 
-# Ide generated files
-.kdev4/*
-*.kdev4
 *.swp
 .DS_Store
 
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,9 @@ public/minjs/
 public/js/main.js
 Gemfile.lock
 
+# Ide generated files
+.kdev4/*
+*.kdev4
 *.swp
 .DS_Store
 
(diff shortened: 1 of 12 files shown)


jonathanverner:
Hmm..., I completely missed writing (and modifying) the unit tests  :-( 

Before I start working on them, though, could you tell me whether its worth the effort, i.e. I don't want to spend too much effort on writing tests for code that has no chance of being accepted anyway...

henryoswald:
Hi Jonathan,

We are not going to be adding public registration into the community edition I am afraid, feel free to maintain a fork though!

jonathanverner:
Hi Henry, 

no problem. And again, thanks for sharing the community edition as open source! Its a great piece of software.

henryoswald:
Thanks for being understanding Jonathan, we didn't fully anticipate how hard it would be to keep everyone happy.

PR #187, opened 2015-05-28
Fix call to addUserToProject.

dwrensha:
I was working on permissions support in the Sandstorm port, and I noticed that `changeUsersPriviliegeLevel()` does not quite work. This patch seems to fix things.

I ended up not actually needing to use `changeUsersPrivilegeLevel()`, and it looks like it's currently not used anywhere in this repo, so another option might be simply to delete the function.


henryoswald:
thanks for pointing it out, i've deleted the function now if it is isn't being used - https://github.com/sharelatex/web-sharelatex/commit/1b4f09bf9f11ff0adf171152d878ab85a239142f

PR #206, opened 2015-11-09
Updated package warning parser

mathsguy42:
Added a single function to handle all package warnings, either one-line or multi-line, and finding the correct line number if any. The function takes advantage of the fact that LaTeX inserts a blank line after each warning.

The main changes are:

* Modified the parser to include the string **Package (packagename) Warning:**, this makes it easier to figure out what is causing the warning, specially when there's no line number.

* Added tests for the *geometry* package.

* Added tests for the *caption* package. Warnings from this one are hard to parse because the second line has a *(caption)* prefix and also includes the source-file line number, and the last line doesn't have blank spaces at the beginning.


henryoswald:
LGTM, @jpallen ?

jpallen:
Yes, sorry for the delay. This is just an update of the patch that has already been applied to jpallen/latex-log-parser. LGTM

PR #210, opened 2015-11-15
Replaces '&' with '\&' when creating a project

rrdelaney:
When creating a project, if the user puts a '&' in the project title, the project's initial build will fail because 

``` latex
\title{This & That}
```

is not valid. 

To fix this, the title must be 

```
\title{This \& That}
```

### DIFF
--- a/app/coffee/Features/Project/ProjectCreationHandler.coffee
+++ b/app/coffee/Features/Project/ProjectCreationHandler.coffee
@@ -30,7 +30,7 @@ module.exports =
 		self = @
 		@createBlankProject owner_id, projectName, (error, project)->
 			return callback(error) if error?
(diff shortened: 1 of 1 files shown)


rrdelaney:
Yep, I'll implement those too

rrdelaney:
This should be good now

henryoswald:
Thanks, I need to check if this will effect how our drobox sync works before merging.

henryoswald:
@jpallen I think this should actually be fine to go as is.

rrdelaney:
Any update on this?

jpallen:
Sorry, this got a bit overlooked. This looks like a good change. If it had a unit test, we'd be happy to merge it.

PR #212, opened 2015-11-24
[WIP] Adding public API v1

danielneis:
Hello,

this adds the first version of the public API that exposes some routes to be accessed using the existing auth_token for users.
This is intended primarily to work in the integration with OSF software, but may be used with other projects.
This work was started after reading of #299 and some conversations with Henry and James.

Some unit test are failing but we were unable to understando what is causing the error.
Here is the link to the test ouput:
https://gist.github.com/danielneis/0c13e794a80b886a5f1e
Any help woul be appreciated.

Kind regards,
Daniel

### DIFF
--- a/app/coffee/Features/FileStore/FileStoreController.coffee
+++ b/app/coffee/Features/FileStore/FileStoreController.coffee
@@ -15,9 +15,9 @@ is_html = (file) ->
 
 	ends_with('.html') or ends_with('.htm') or ends_with('.xhtml')
 
-module.exports =
+module.exports = FileStoreController =
 
-	getFile : (req, res)->
+	getFile: (req, res)->
 		project_id = req.params.Project_id
 		file_id = req.params.File_id
 		queryString = req.query
@@ -37,3 +37,12 @@ module.exports =
 					res.setHeader('Content-Type', 'text/plain')
 				res.setHeader("Content-Disposition", "attachment; filename=#{file.name}")
 				stream.pipe res
+
+	getFileByProjectAndPath: (req, res)->
+		project_id = req.params.project_id
+		path = req.params.path
+		ProjectLocator.findElementByPath project_id, path, (err, foundFolder) ->
+			if foundFolder?
+                            req.params.File_id = foundFolder._id
+                            req.params.Project_id = project_id
+                            FileStoreController.getFile(req, res)
--- a/app/coffee/Features/FileStore/FileStoreController.coffee
+++ b/app/coffee/Features/FileStore/FileStoreController.coffee
@@ -67,3 +67,5 @@ module.exports = FileStoreController =
 								res.setHeader('Content-Type', 'text/plain')
 							res.setHeader("Content-Disposition", "attachment; filename=#{file.name}")
 							stream.pipe res
+			logger.err err:err, project_id: project_id, path: path, queryString:queryString, "error finding file by project and path"
+			return res.sendStatus 500
--- a/app/coffee/Features/FileStore/FileStoreController.coffee
+++ b/app/coffee/Features/FileStore/FileStoreController.coffee
@@ -67,5 +67,6 @@ module.exports = FileStoreController =
 								res.setHeader('Content-Type', 'text/plain')
 							res.setHeader("Content-Disposition", "attachment; filename=#{file.name}")
 							stream.pipe res
-			logger.err err:err, project_id: project_id, path: path, queryString:queryString, "error finding file by project and path"
-			return res.sendStatus 500
+			else
+				logger.err err:err, project_id: project_id, path: path, queryString:queryString, "error finding file by project and path"
+				return res.sendStatus 500
--- a/app/coffee/Features/FileStore/FileStoreController.coffee
+++ b/app/coffee/Features/FileStore/FileStoreController.coffee
@@ -1,6 +1,7 @@
 logger = require('logger-sharelatex')
 FileStoreHandler = require("./FileStoreHandler")
 ProjectLocator = require("../Project/ProjectLocator")
+ProjectEntityHandler = require("../Project/ProjectEntityHandler")
 _ = require('underscore')
 
 is_mobile_safari = (user_agent) ->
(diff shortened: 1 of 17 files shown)


app/coffee/Features/FileStore/FileStoreController.coffee
```diff
@@ -37,3 +38,32 @@ module.exports =
...
 				res.setHeader("Content-Disposition", "attachment; filename=#{file.name}")
 				stream.pipe res
+
+	getFileByProjectAndPath: (req, res)->
+		project_id = req.params.project_id
+		path = req.params.path
+		queryString = req.query
+		user_agent = req.get('User-Agent')
+		ProjectLocator.findElementByPath project_id, path, (err, foundFolder) ->
+			if foundFolder?
```
henryoswald: what happens if there is no found folder?

app/coffee/Features/Project/ProjectController.coffee
```diff
@@ -289,6 +291,29 @@ module.exports = ProjectController =
 					project.owner = users[project.owner_ref.toString()]
 			callback null, projects
 
+	getProjectList: (req, res, next)->
```
henryoswald: add tests for these 2 functions into ProjectControllerTests.coffee please?

app/coffee/models/Doc.coffee
```diff
@@ -5,7 +5,8 @@ Schema = mongoose.Schema
 ObjectId = Schema.ObjectId
 
 DocSchema = new Schema
-	name          :     {type:String, default:'new doc'}
+	name          :     {type:String, default:'new doc'},
+        mimetype	:	{type:String, default:'application/x-tex'}
```
henryoswald: we won't want to add this mimetype into the document ideally, does it ever change? If not can you just use mime.lookup(name) when it is required

app/coffee/models/File.coffee
```diff
@@ -7,7 +8,8 @@ ObjectId = Schema.ObjectId
 FileSchema = new Schema
 	name            :     type:String, default:''
 	created     	:     type:Date, default: () -> new Date()
-	rev 			:	  {type:Number, default:0}
+	rev 			:	  {type:Number, default:0},
+        mimetype	:	type:String, default: () -> mime.lookup(name)
```
henryoswald: same as doc that we won't want to store this in the db

henryoswald:
In FileStoreControllerTests try adding a stub of Project Entity Handler like below

			"../Project/ProjectLocator": @ProjectLocator
			"../Project/ProjectEntityHandler":{}
			"./FileStoreHandler": @FileStoreHandler


Also we will need good unit tests on functions like getFileByProjectAndPath before we can think about merging.

danielneis:
Hello, @henryoswald 

we've made some changes to address the things you write about on your last review.
Would you please take a second look at the code? It should be quite finished now =)

Kind regards,
Daniel

danielneis:
hello, @henryoswald how are things going?
it would be nice if you guys could review this =)

app/coffee/Features/Project/ProjectEditorHandler.coffee
```diff
@@ -65,9 +66,11 @@ module.exports = ProjectEditorHandler =
 		docs       : @buildDocModelView doc for doc in folder.docs
 
 	buildFileModelView: (file) ->
-		_id  : file._id
-		name : file.name
+		_id      : file._id
+		name     : file.name
+		mimetype : mime.lookup(name)
```
felliott: Should this be `mime.lookup(file.name)`?

henryoswald:
Hey,

sorry for no reply. The code is looking better. Does it do the job for you? TBH we are still a long way away from having an api on ShareLaTeX.com or merging this in as the hard part of auth still needs sorting.

app/coffee/Features/FileStore/FileStoreController.coffee
```diff
@@ -37,3 +38,34 @@ module.exports =
...
+							if err? or !file?
+								logger.err err:err, project_id: project_id, file_id: file_id, queryString:queryString, "error getting file stream for downloading file"
+								return res.sendStatus 500
+							# mobile safari will try to render html files, prevent this
+							if (is_mobile_safari(user_agent) and is_html(file))
+								logger.log filename: file.name, user_agent: user_agent, "sending html file to mobile-safari as plain text"
+								res.setHeader('Content-Type', 'text/plain')
+							res.setHeader("Content-Disposition", "attachment; filename=#{file.name}")
+							stream.pipe res
+			logger.err err:err, project_id: project_id, path: path, queryString:queryString, "error finding file by project and path"
```
felliott: I don't speak fluent coffeescript, so I might be missing an implicit return, but I think this should be inside an else condition.  With it in, I get a 500 when I try to download my `.tex` file.  Without it, it downloads fine.

danielneis:
Hello, 

we have fixed the last bits here and rebased the code to current master.

Could you please elaborate on "the hard part of auth still needs sorting" ?

We have discussed before about that and the only thing I can remember is implement the OAuth2 protocal that in essence is the automatic sharing of tokens between apps much like what we are doing here using the user auth token. It seems to be no security risks in the current approach so I would not make that a real barrier to get the work integrated.

Best regards,
Daniel

henryoswald:
Hi Daniel,

the oauth2 also includes things like permissions management for project access. We will need pages and mechanisms to allow developer x to access project y which is a complex bit of work unfortunately. It will need things like rate limiting, documentation and to be easy for us to manage if we release it on ShareLaTeX.com

felliott:
Hey @henryoswald!

My name is Fitz Elliott and I'm a developer on the OSF.  We've had a few users inquire about being able to connect their ShareLaTeX accounts to their OSF accounts.  They're big fans of your service and would love to be able to write their manuscripts on your site and have them archived on ours.  Do you have an estimation of when a public API for ShareLaTeX will be available?  @danielneis and his team have already written the integration code, so we'd be able to go live and announce pretty soon after you release.  Thank you for your time and have a great day!

Cheers,
Fitz (@felliott)

henryoswald:
Hi Fitz,

thanks for the comment. We are really quite a long way away from an API yet im afraid. Our current workload on ShareLaTeX is pretty high and its going to be hard for us to fit this in. Although the endpoints are written the more complex stuff around authentication and how we expose an api need to be worked out. If we are going to support it on ShareLaTeX then we really feel we have todo a good job of something we want to support for years to come.

danielneis:
Hello,

I was looking at the GitLab public API, and have found that they have the same feature for authenticating using a private token as you can see on this link: http://docs.gitlab.com/ce/api/

Maybe you should tell them how insecure their project is ;)

henryoswald:
Hey Daniel, 

The principal of private tokens is not bad, but the ones we use were not designed with this in mind. We also have to deal with refreshing them, permissions access, documentation, rate limiting etc which is why we can't just release this yet.

SijmenSchoon:
Is there any news on this?

PR #219, opened 2016-01-19
An institution-wise OAuth replacement for login/registration

niedakh:
During our work at the ENGINE [1] project - one of the tasks we have at hand is to set up a platform for researchers in the centre with a unified oauth-based login to all parts of the locally hostem platform. We've wanted to be able to use sharelatex with that, so I've added an option to login via oauth2 instead of normal login approach. 

Would you be interested in such a functionality being merged? If so - please comment on what your expectations would be for it to be merged in - should I clean up the code some more, are there any coding guidelines for sharelatex?

What I'm sending here is just my first working version, if oauth enabled in settings it replaces the login form with an oauth login button and then performs the login operation - creating the user and logging in. I'm willing to polish this code some more to your liking, if you feel this kind of functionality is welcome in shareltex. 

[1] http://engine.pwr.edu.pl

### DIFF
--- a/app/coffee/Features/Authentication/OAuthController.coffee
+++ b/app/coffee/Features/Authentication/OAuthController.coffee
@@ -1,0 +1,105 @@
+AuthenticationManager = require ("./AuthenticationManager")
+LoginRateLimiter = require("../Security/LoginRateLimiter")
+
+User = require("../../models/User").User
+UserGetter = require "../User/UserGetter"
+UserUpdater = require "../User/UserUpdater"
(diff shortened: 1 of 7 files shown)


henryoswald:
Hi,

thanks for this PR. Which oauth provider has it been designed to work with?

niedakh:
It complies to http://projects.spring.io/spring-security-oauth/docs/oauth2.html - we've also written an oauth management solution in spring that we will opensource very soon.

PR #229, opened 2016-03-03
add client-side spelling cache to reduce load on server

briangough:
cache the last 1,000 words so that updates when typing will mostly be cached (the same line is sent to the server each time, with the addition of the latest word typed).


jpallen:
Was just reading your branch :) LGTM :+1:

PR #235, opened 2016-04-04
Correct alignment for word count dialog

mdamien:
The word count window does the wrong alignment (as you can see in the picture) if the window width is less than the bootstrap `md` size (992px).

image


henryoswald:
Thanks!

PR #243, opened 2016-05-07
Moved fonts to localhost for offline usage compatibility

mghadam:


henryoswald:
Thanks, we might merge this in the near future once we get a cnd sorted for ShareLaTeX.com, at the moment I don't want to put more load on our servers.

mghadam:
I've made it for sharelatex docker image in fact.

henryoswald:
Thanks, Im aware of that but anything done here goes up to ShareLaTeX.com and other hosted versions as well.

henryoswald:
I don't think we are ever going to serve these locally sorry.

clemh78:
I have same problem. I want to host sharelatex in offline network but the fonts can not load.

Is it possible to plan this feature for docker version ?

PR #275, opened 2016-07-08
Track positive and negative feedback as different events

paulojreis:
Track positive and negative feedback as different events, for easier analysis.


jpallen:
LGTM

PR #286, opened 2016-07-19
fix close editor button

henryoswald:
- only evaulate close on web router, not api/static assets
- allow /admin pages to still be available

### DIFF
--- a/app/coffee/infrastructure/Server.coffee
+++ b/app/coffee/infrastructure/Server.coffee
@@ -101,19 +101,19 @@ if app.get('env') == 'production'
 	logger.info "Production Enviroment"
 	app.enable('view cache')
 
-
-
 app.use (req, res, next)->
 	metrics.inc "http-request"
 	crawlerLogger.log(req)
 	next()
 
-app.use (req, res, next) ->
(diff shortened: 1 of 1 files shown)


app/coffee/infrastructure/Server.coffee
```diff
@@ -101,19 +101,19 @@ if app.get('env') == 'production'
...
 	metrics.inc "http-request"
 	crawlerLogger.log(req)
 	next()
 
-app.use (req, res, next) ->
-	if !Settings.editorIsOpen
+webRouter.use (req, res, next) ->
+	if Settings.editorIsOpen
+		next()
+	else if req.url.indexOf("/admin") == 0
```
jpallen: Make sure /admin is only at the start, otherwise we might get caught out by an URL like /project/:id/collaborators/add/admin or something

jpallen: Oh nm, I see that's what indexOf == 0 does. Maybe should be more explicit? Maybe it's just late and I can't process code.

henryoswald: == 0 should do that no?

PR #289, opened 2016-07-27
Countly pruning

paulojreis:
Diminish the amount of data being sent to Countly:

- Don't track page views
- Don't track sessions
- Track a few events only the first time they happen for a given user (we only want to know it happened once, not the amount/frequency).

### DIFF
--- a/public/coffee/ide.coffee
+++ b/public/coffee/ide.coffee
@@ -72,13 +72,13 @@ define [
 		# Tracking code.
 		$scope.$watch "ui.view", (newView, oldView) ->
 			if newView? and newView != "editor" and newView != "pdf"
-				event_tracking.sendCountly "ide-open-view-#{ newView }" 
+				event_tracking.sendCountlyOnce "ide-open-view-#{ newView }" 
 
 		$scope.$watch "ui.chatOpen", (isOpen) ->
-			event_tracking.sendCountly "ide-open-chat" if isOpen
+			event_tracking.sendCountlyOnce "ide-open-chat" if isOpen
 
 		$scope.$watch "ui.leftMenuShown", (isOpen) ->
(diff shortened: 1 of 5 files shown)


public/coffee/ide.coffee
```diff
@@ -72,13 +72,13 @@ define [
 		# Tracking code.
 		$scope.$watch "ui.view", (newView, oldView) ->
 			if newView? and newView != "editor" and newView != "pdf"
-				event_tracking.sendCountly "ide-open-view-#{ newView }" 
+				event_tracking.sendCountlyOnce "ide-open-view-#{ newView }"
```
henryoswald: shall we change these event names to something like "has-done-ide-open-view" or "onetime-ide-open-view"? open for debate.

public/coffee/main/event.coffee
```diff
@@ -1,6 +1,39 @@
 define [
 	"base"
 ], (App) ->
+	CACHE_KEY = "countlyEvents"
+
+	_getEventCache = () -> 
+		eventCacheStr = window.localStorage.getItem CACHE_KEY
```
henryoswald: we have modules/localStorage which wraps this up a little bit prevents some errors (try catch)

public/coffee/main/event.coffee
```diff
@@ -1,19 +1,50 @@
...
+			if !eventCache?
+				eventCache = {}
+				localStorage CACHE_KEY, eventCache 
+
+			return eventCache
+
+		_eventInCache = (key) ->
+			curCache = _getEventCache()
+
+			if (curCache.hasOwnProperty key)
```
henryoswald: out of interest what does hasOwnProperty add over doing

curCache[key] || false

PR #304, opened 2016-08-23
Ho jade speedup

henryoswald:


henryoswald:
playing locally it seems to take ~250-300ms, node_env as production.

Numbers taken from tryRender function in express/lib/application.js

commenting out //- include ./editor/left-menu reduced the time to ~10ms

commenting out parts of left menu showed moduleIncludes is very slow

putting loads of moduleIncludes makes it very slow, moduleIncludesAvailable is not slow.

PR #305, opened 2016-08-23
don't use cdn if it can not be accessed

henryoswald:


app/coffee/infrastructure/ExpressLocals.coffee
```diff
@@ -61,11 +61,18 @@ module.exports = (app, webRouter, apiRouter)->
 
 	webRouter.use (req, res, next)-> 
 
+		cdnBlocked = req.query.nocdn == 'true' or req.session.cdnBlocked
+
+		if cdnBlocked and !req.session.cdnBlocked?
+			ip = req.ip || req.socket?.socket?.remoteAddress || req.socket?.remoteAddress
```
jpallen: I think express has a clever ip method that looks at forwarded headers, etc. req.ip will always be the load balancer I think

app/views/layout.jade
```diff
@@ -51,8 +52,16 @@ html(itemscope, itemtype='http://schema.org/Product')
 			window.csrfToken = "#{csrfToken}";
 			
 		block scripts
-		script(src=buildJsPath("libs/jquery-1.11.1.min.js", {fingerprint:false}))
 		script(src=buildJsPath("libs/angular-1.3.15.min.js", {fingerprint:false}))
+		script(src=buildJsPath("libs/jquery-1.11.1.min.js", {fingerprint:false}))
```
jpallen: Any reason to load jquery after angular, or is it just an accidental move? I'd move it back if not, just in case it matters.

app/coffee/infrastructure/ExpressLocals.coffee
henryoswald: i stole it from http.coffee in metrics where do this, assumed there are some edge cases.

jpallen: It seems req.ip is the right thing to use. http://expressjs.com/en/api.html#req.ip Other code is left over from socket.io, so dead.

jpallen: Hmm, that's a lie, I think it's from old express. Anyway, to keep our code up to date and clean, I think it should just be `req.ip`.

PR #320, opened 2016-09-21
Don't allow files to be created with / or * characters

jpallen:
Requires the following translations:

  "files_cannot_include_invalid_characters": "Files cannot include '*' or '/' characters",
  "invalid_file_name": "Invalid File Name"

### DIFF
--- a/public/coffee/ide/file-tree/controllers/FileTreeEntityController.coffee
+++ b/public/coffee/ide/file-tree/controllers/FileTreeEntityController.coffee
@@ -26,9 +26,23 @@ define [
 		$scope.startRenaming = () ->
 			$scope.entity.renaming = true
 
+		invalidModalShowing = false
 		$scope.finishRenaming = () ->
-			delete $scope.entity.renaming
 			name = $scope.inputs.name
+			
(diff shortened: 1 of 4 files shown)


public/coffee/ide/file-tree/controllers/FileTreeEntityController.coffee
```diff
@@ -26,9 +26,23 @@ define [
 		$scope.startRenaming = () ->
 			$scope.entity.renaming = true
 
+		invalidModalShowing = false
 		$scope.finishRenaming = () ->
-			delete $scope.entity.renaming
 			name = $scope.inputs.name
+			
+			if !name.match(/^[^\*\/]*$/)
```
ShaneKilkelly: Should this regex not be a reference to `ide.validFileRegex`?

jpallen: Yes, it should, thanks

jpallen: Fixed now

PR #333, opened 2016-10-05
Support package versions

briangough:
Put all client package versions (pdfjs, moment and in future, ace) into a module.


jpallen:
LGTM

PR #351, opened 2016-10-26
Ask for password when deleting user account

ShaneKilkelly:


app/coffee/Features/User/UserController.coffee
```diff
@@ -15,13 +15,35 @@ settings = require "settings-sharelatex"
 
 module.exports = UserController =
 
+	# TODO: deprecated, remove
```
jpallen: Any reason to not just remove now? I think the risk of people on the account page, but not hit 'delete' yet is small.

ShaneKilkelly: True, will remove

paulojreis:
Looks good.

PR #353, opened 2016-10-27
AB test history teaser

paulojreis:
Adds an AB test for the history feature teaser, much like the previous Dropbox one. Also adds Metabase tracking for both.


jpallen:
LGTM 👍

PR #361, opened 2016-11-02
Pass ignoreTLS option to nodemailer

djwf:
This fixed a problem I was having with sending password reset emails via SMTP port 25 on a server that supports STARTTLS, but on which I did not want to use STARTTLS.

Without this change I would get a "CERT_UNTRUSTED" error, and nodemailer never sees the ignoreTLS option set in the environment.


henryoswald:
Thanks for this.

PR #363, opened 2016-11-03
lock editor when connection down

briangough:
block access to the editor after 15 seconds of unsaved changes.

reconnect attempts continue in the background.


jpallen:
LGTM

PR #373, opened 2016-11-17
Show email on settings when using external auth

ShaneKilkelly:
Also defend server-side against the email address being set, when using an external auth source.


app/coffee/Features/User/UserController.coffee
```diff
@@ -44,6 +44,7 @@ module.exports = UserController =
 
 	updateUserSettings : (req, res)->
 		user_id = AuthenticationController.getLoggedInUserId(req)
+		usingExternalAuth = !!(settings.ldap? or settings.saml?)
```
ShaneKilkelly: Possibly not, just forcefully coercing to boolean. Without the `!!`, `usingExternalAuth` ends up as whatever the value of `settings.saml` or `settings.ldap` is. It'll be truthy for sure, but doesn't feel as clean.

ShaneKilkelly: Actually, no, disregard, the `?` takes care of that, will remove `!!`

henryoswald: wouldn't this always result in a true or false?

usingExternalAuth = settings.ldap? or settings.saml?

henryoswald: maybe we should add a settings.usingExternalAuth = settings.ldap? or settings.saml?

PR #383, opened 2016-12-06
allow user to delete account with cancelled subscription

ShaneKilkelly:
Ignore the "Can't transition from cancelled to cancelled" error from Recurly.


app/coffee/Features/Subscription/RecurlyWrapper.coffee
```diff
@@ -418,6 +418,9 @@ module.exports = RecurlyWrapper =
 			url: "subscriptions/#{subscriptionId}/cancel",
 			method: "put"
 		}, (error, response, body) ->
+			if error? and body?.match(/.*A canceled subscription can't transition to canceled.*/)
```
henryoswald: do they not give a nice body xml error.message thing?

PR #388, opened 2016-12-19
Sk rate limit cluster

ShaneKilkelly:
- Make it possible to use redis-cluster for rate-limiting
- Remove dependency on `redback`, centralize our own rate-limiting logic.
- Replace with a fork of `rolling-rate-limiter`, with added `ioredis` support (https://github.com/ShaneKilkelly/rolling-rate-limiter)
- Tidy up stray redis imports in modules where the imports were not used


app/coffee/infrastructure/RateLimiter.coffee
```diff
@@ -1,15 +1,27 @@
...
+module.exports = RateLimiter =
+
+	_buildKey: (endpoint, subject) ->
+		return "RateLimiter:#{endpoint}:{#{subject}}"
+
+	addCount: (opts, callback = (err, shouldProcess)->)->
+		k = RateLimiter._buildKey(opts.endpointName, opts.subjectName)
+		multi = rclient.multi()
+		multi.incr(k)
+		multi.get(k)
```
henryoswald: I guess with this system if a user tries the operation again it will reset the ttl and not expire so if a user tries to login 8 times in min 1, 3 times in min 2 and hits the rate limit they could try every minutes forever and it will still be on the rate limit. Is that how redback worked?

henryoswald:
LGTM, do you want to create a PR to classdojo/rolling-rate-limiter so we can use npm module?

henryoswald:
shall we get this out on monday?

PR #395, opened 2017-01-20
Sk launchpad

ShaneKilkelly:
- Add a new 'testEmail' email type
- add `on-success` and `on-error` optional callbacks to async-form

### DIFF
--- a/public/coffee/directives/asyncForm.coffee
+++ b/public/coffee/directives/asyncForm.coffee
@@ -33,6 +33,10 @@ define [
 							response.success = true
 							response.error = false
 
+							if onSuccessHandler = scope[attrs.onSuccess]
+								onSuccessHandler(data, status, headers, config)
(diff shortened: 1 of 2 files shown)


public/coffee/directives/asyncForm.coffee
```diff
@@ -33,6 +33,10 @@ define [
 							response.success = true
 							response.error = false
 
+							if onSuccessHandler = scope[attrs.onSuccess]
+								onSuccessHandler(data, status, headers, config)
```
henryoswald: is this 'if' shorthand for

    onSuccessHandler = scope[attrs.onSuccess]
    if onSuccessHandler?
        dostuf()
   
if so its a touch unusual imo

ShaneKilkelly: Fair, will change

PR #399, opened 2017-01-25
promote domain alerts

henryoswald:


app/coffee/Features/Announcements/AnnouncementsHandler.coffee
```diff
@@ -5,20 +5,44 @@ _ = require("lodash")
...
+				return email.indexOf(domain) != -1
+			return matches.length > 0
+		return domainSpecific or []
+
+
+	getUnreadAnnouncements : (user, callback = (err, announcements)->)->
 		async.parallel {
 			lastEvent: (cb)->
-				AnalyticsManager.getLastOccurance user_id, "announcement-alert-dismissed", cb
+				AnalyticsManager.getLastOccurance user?._id, "announcement-alert-dismissed", cb
```
jpallen: At this point, if the user doesn't exist we're presumably not going to anything sensible from this? So perhaps just a top level null user guard that bails out with an error, rather than potentially failing silently?

app/coffee/Features/Announcements/AnnouncementsHandler.coffee
```diff
@@ -5,20 +5,44 @@ _ = require("lodash")
...
 
-	getUnreadAnnouncements : (user_id, callback = (err, announcements)->)->
+
+
+module.exports = AnnouncementsHandler = 
+
+
+
+	_domainSpecificAnnouncements : (email)->
+		domainSpecific = _.filter settings?.domainAnnouncmentsToShow, (domainAnnouncment)->
```
jpallen: domainAnnouncmentsToShow -> domainAnnounc**e**mentsToShow. Wouldn't normally care, but this is a settings file entry

app/coffee/Features/Announcements/AnnouncementsHandler.coffee
```diff
@@ -5,20 +5,44 @@ _ = require("lodash")
...
 			announcements: (cb)->
 				BlogHandler.getLatestAnnouncements cb
 		}, (err, results)->
 			if err?
-				logger.err err:err, user_id:user_id, "error getting unread announcements"
+				logger.err err:err, user_id:user?._id, "error getting unread announcements"
 				return callback(err)
 			
-			announcements = _.sortBy(results.announcements, "date").reverse()
+			domainSpecific = AnnouncementsHandler._domainSpecificAnnouncements(user?.email)
```
jpallen: Worth adding some sort of error check that domain announcements have an id? If they don't they could end up impossible to dismiss? Or somewhat to enforce the schema through the settings file? Not essential, just thinking of ways this could get screwed up by accident in future.

PR #421, opened 2017-02-14
mark announcments before user signs up as read

henryoswald:


app/coffee/Features/Announcements/AnnouncementsHandler.coffee
```diff
@@ -19,6 +19,9 @@ module.exports = AnnouncementsHandler =
 		if !user? and !user._id?
 			return callback("user not supplied")
 
+		timestamp = user._id.toString().substring(0,8)
+		userSignupDate = new Date( parseInt( timestamp, 16 ) * 1000 )
```
jpallen: We have an explicit signUpDate attribute we could use here?

app/coffee/Features/Announcements/AnnouncementsHandler.coffee
```diff
@@ -48,7 +51,9 @@ module.exports = AnnouncementsHandler =
 				announcement.id == lastSeenBlogId
 
 			announcements = _.map announcements, (announcement, index)->
-				if announcementIndex == -1
+				if announcement.date < userSignupDate
```
jpallen: Is announcement.date always midnight of that day? so a user signing up at any time on the day of the announcement won't see it? Just checking it isn't the other way around, and user signing up on the same day as the annoucement will always see it due to the 24 hour window of a day.

app/coffee/Features/Announcements/AnnouncementsHandler.coffee
henryoswald: This user is from the session. Wanted to avoid another call to db while this is via http and isolated.

app/coffee/Features/Announcements/AnnouncementsHandler.coffee
henryoswald: yes it is always midnight

PR #424, opened 2017-02-15
Give the scrollbar a background-color, so that Safari knows that it s…

paulojreis:
…hould paint it.

Fixes invisible scrollbars in Safari, as per https://sharelatex-accounts.groovehq.com/groove_client/mailboxes/33479/filters/183581/tickets/38653216

Issue seems to come from Ace + Safari (Safari probably being the biggest culprit here, when computing non-paintable regions). Solution (and issue) from https://github.com/ajaxorg/ace/issues/2872

### DIFF
--- a/public/stylesheets/app/editor.less
+++ b/public/stylesheets/app/editor.less
@@ -181,6 +181,19 @@
 	}
 }
 
+// Hack to solve an issue where scrollbars aren't visible in Safari.
+// Safari seems to clip part of the scrollbar element. By giving the 
+// element a background, we're telling Safari that it *really* needs to 
+// paint the whole area. See https://github.com/ajaxorg/ace/issues/2872
+.ace_scrollbar-inner {
+	background-color: #FFF;
+	opacity: 0.01;
+
+	.ace_dark & {
+		background-color: #000;
+	}
+}
+
 .ui-layout-resizer {
 	width: 6px;
 	background-color: #f4f4f4;


Exadra37:
Just for anyone arriving here(like me) from Google to fix this same issue in their own project the solution can be found in [this site](https://www.filamentgroup.com/lab/scrollbars/):

```
:root {
  --scrollbar-track-color: transparent;
  --scrollbar-color: rgba(0,0,0,.2);

  --scrollbar-size: .375rem;
  --scrollbar-minlength: 1.5rem; /* Minimum length of scrollbar thumb (width of horizontal, height of vertical) */
}
.overflowing-element::-webkit-scrollbar {
  height: var(--scrollbar-size);
  width: var(--scrollbar-size);
}
.overflowing-element::-webkit-scrollbar-track {
  background-color: var(--scrollbar-track-color);
}
.overflowing-element::-webkit-scrollbar-thumb {
  background-color: var(--scrollbar-color);
  /* Add :hover, :active as needed */
}
.overflowing-element::-webkit-scrollbar-thumb:vertical {
  min-height: var(--scrollbar-minlength);
}
.overflowing-element::-webkit-scrollbar-thumb:horizontal {
  min-width: var(--scrollbar-minlength);
}
```

Just replace `.overflowing-element` with the class name for the div you want to show the scroll bar.

**NOTE:** Sorry to hijack this issue, but this problem is so shard to solve and find a solution, because Google keeps directing me to issues in Github for specific projects or to tutorials in how to enable the scroll bars to always show in the MAC OS.

PR #429, opened 2017-02-17
Collapse overview entries

paulojreis:
Adds expand/collapse behaviour per file in the overview panel. Also adds the number of entries, when a file is collapsed.


jpallen:
As discussed, make collapse indicator consistent with file tree, and persist state between sessions.

public/coffee/ide/review-panel/controllers/ReviewPanelController.coffee
```diff
@@ -27,6 +28,16 @@ define [
 			rendererData: {}
 			loadingThreads: false
 
+		$scope.$on "project:joined", () ->
+			$scope.reviewPanel.overview.docsCollapsedState = JSON.parse(localStorage("docs_collapsed_state:#{$scope.project_id}")) or {}
```
jpallen: Doing this on join project will reset it if the client loses connection and reconnects. Just do it once on load, i.e. when the controller is created?

public/coffee/ide/review-panel/controllers/ReviewPanelController.coffee
```diff
@@ -27,6 +28,16 @@ define [
...
+
+		window.addEventListener "beforeunload", () ->
+			collapsedStates = {}
+			for doc, state of $scope.reviewPanel.overview.docsCollapsedState
+				collapsedStates[doc] = state if state is true
```
jpallen: I find this inline syntax hard to parse and never trust it. Maybe make it explicit?

PR #437, opened 2017-02-23
Change event handling to dirty state, and change behaviour if it's an undo op

jpallen:


jpallen:
There are two different changes that ended up in one PR here, sorry:

1) Migrate away from an event machine to tracking a dirty state. See commit at  https://bitbucket.org/sharelatex/track-changes-engine/commits/4dc817c74cc306d581b345461864128f57a4ec55 for the RangeTracker changes for this bit, and the changes in https://github.com/sharelatex/web-sharelatex/pull/437/files#diff-7f4269502616015fb6a887b3ab7ef049 for the front-end UI changes.

2) Changing how we handle inserts next to delete ranges depending whether it was an undo/revert, or just a normal insert. Note that the changes to the sharejs engine (text.coffee) are unit tested in the doc updater in https://github.com/sharelatex/document-updater-sharelatex/pull/19

PR #451, opened 2017-03-09
When upload to filestore fails, produce an error

ShaneKilkelly:


app/coffee/Features/FileStore/FileStoreHandler.coffee
```diff
@@ -25,10 +25,19 @@ module.exports = FileStoreHandler =
 				timeout:fiveMinsInMs
 			writeStream = request(opts)
 			readStream.pipe writeStream
-			writeStream.on "end", callback
+
+			writeStream.on 'response', (response) ->
+				if response.statusCode not in [200, 201]
```
jpallen: I often use the pattern 200 <= statusCode < 300 to catch any 'success' code, but I think this is fine here.

jpallen:
@ShaneKilkelly This LGTM, assuming if this new code errors, it doesn't put the file object in the project still?

app/coffee/Features/FileStore/FileStoreHandler.coffee
jpallen: Ah, I see it's hard to match on a 'not' basis. I think I'm just bikeshedding code style :), this LGTM

PR #456, opened 2017-03-17
Show an explanation error message on too many track changes error

jpallen:


public/coffee/ide/editor/EditorManager.coffee
```diff
@@ -121,11 +121,22 @@ define [
...
+				else if typeof error == "string"
+					message = error
+				else
+					message = ""
+				if message.match "maxDocLength"
 					@ide.showGenericMessageModal(
 						"Document Too Long"
 						"Sorry, this file is too long to be edited manually. Please upload it directly."
 					)
+				else if message.match "too many comments or tracked changes"
```
henryoswald: bit strange sending a sentence down rather than something like maxRangesHit, assume its a pain to refactor?

jpallen: Yea, I didn't like this, but it's what the doc updater currently returns. The error above, with a nicer message is emitted internally to the client, so we have more direct control. My default position was to not change out of laziness, but I will do if you ask me to.

henryoswald: im lazy. LGTM

PR #458, opened 2017-03-20
Add option to hard-hide track-changes onboarding.

ShaneKilkelly:
This is useful for on-prem installations where Analytics
services is not operating. Previously the Analytics call
would be a no-op, but that would mean the prompt would be
shown every time a user loaded a project.

It's necessary to have a way to perma-hide this prompt
via settings.


ShaneKilkelly:
Irrelevant, taking another approach in https://github.com/sharelatex/web-sharelatex/pull/459

PR #465, opened 2017-03-27
If sending email fails, return a generic error.

ShaneKilkelly:
This prevents us from leaking juicy details of our aws/ses setup via the
password-reset form.


app/coffee/Features/Email/EmailSender.coffee
```diff
@@ -74,4 +74,4 @@ module.exports =
 					logger.err err:err, "error sending message"
 				else
 					logger.log "Message sent to #{options.to}"
-				callback(err)
+				callback(new Error('Cannot send email'))
```
jpallen: This is always going to return an error now?

PR #466, opened 2017-03-28
fix unhandled exception in ProjectDetailsHandler

briangough:
if the project does not exist there is an exception when accessing project.owner_ref

### DIFF
--- a/app/coffee/Features/Project/ProjectApiController.coffee
+++ b/app/coffee/Features/Project/ProjectApiController.coffee
@@ -1,4 +1,5 @@
 ProjectDetailsHandler = require("./ProjectDetailsHandler")
+Errors = require("../Errors/Errors")
 logger = require("logger-sharelatex")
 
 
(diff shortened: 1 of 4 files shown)


app/coffee/Features/Project/ProjectDetailsHandler.coffee
```diff
@@ -10,7 +10,7 @@ module.exports =
 
 	getDetails: (project_id, callback)->
 		ProjectGetter.getProject project_id, {name:true, description:true, compiler:true, features:true, owner_ref:true}, (err, project)->
-			if err?
+			if err? or !project?
```
jpallen: If you just want a quick fix, this is fine. But the 'correct' way to do it would be include './Errors' and return a NotFoundError, which will send back a 404 from the HTTP request, and not show up in Sentry etc.

jpallen: https://github.com/sharelatex/web-sharelatex/blob/master/app/coffee/Features/Errors/Errors.coffee

briangough: Have extended it to do a 404 error on NotFound

app/coffee/Features/Project/ProjectApiController.coffee
```diff
@@ -7,8 +8,11 @@ module.exports =
 	getProjectDetails : (req, res)->
 		{project_id} = req.params
 		ProjectDetailsHandler.getDetails project_id, (err, projDetails)->
-			if err?
+			if err? and err instanceof Errors.NotFoundError
```
jpallen: This entire error block can just be passed to next(err). Express is set up to handle NotFoundErrors with a 404, or a 500 for other errors.

briangough: Now uses the Error Handler

PR #473, opened 2017-03-31
Sk upgrade metrics

ShaneKilkelly:


jpallen:
Let's remove the Metrics file here too, and just use `metrics-sharelatex` directly.

PR #483, opened 2017-04-11
Continuing flushing document on tpds worker error

jpallen:


henryoswald:
did you check moving/delete rename etc?

jpallen:
Nope

jpallen:
@henryoswald updated now at a layer deeper to always swallow these errors

PR #491, opened 2017-05-01
Project list improvements

paulojreis:
This PR adds some improvements to the projects list page:

* Filtering options are now persisted in local storage
* A new filter for viewing untagged projects was added
* Also added a small "x" button near each tag in the project list, allowing inline tag removal
* Made the whole project row clickable (to select the project)

### DIFF
--- a/app/views/project/list/project-list.pug
+++ b/app/views/project/list/project-list.pug
@@ -140,20 +140,25 @@
 					ng-repeat="project in visibleProjects | orderBy:predicate:reverse",
 					ng-controller="ProjectListItemController"
 				)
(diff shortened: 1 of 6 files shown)


jpallen:
The x is a little off-center, both vertically and horizontally. Can we have the same margin on top, right, and bottom, or is that tricky to specify exactly?

screen shot 2017-05-02 at 10 51 19

paulojreis:
I can fix it vertically. Horizontally it might be harder to "define"; as it is, the spacing between label name and "x" is the same as the overall label horizontal padding.

PR #497, opened 2017-05-11
add cooldown to tpds `mergeUpdate` path

ShaneKilkelly:


app/coffee/Features/ThirdPartyDataStore/TpdsUpdateHandler.coffee
```diff
@@ -24,10 +26,15 @@ module.exports =
 					cb err, project
 		getOrCreateProject (err, project)->
```
henryoswald: This call desearlises the project, negating the performance benefit. Need to kill the request before this point.

henryoswald:
maybe we rate limit on user_id so can still be middlewear?

PR #502, opened 2017-05-12
Don't redirect to images, icons, etc, in login workflow

ShaneKilkelly:


jpallen:
Do we know why the favicon wasnt' caught by the first request? Does the browser request a root asset, i.e. /favicon.ico? How do we know it's a png rather than say ico, and that this will catch it?

jpallen:
Ah, I see it was a touch-icon thing, presumably also caused by this bug where we had a relative path for that, not an absolute one beginning with `/...`? So this is probably already fixed by that fix, but this is another layer of protection?

ShaneKilkelly:
Yup, the favicons, touch-favicons, etc, are in the root directory, not under `/img`.

I don't know if we want to actually move those assets into `/img`, it would be the best solution, but it could be messy with cache-clearing, etc in prod.

PR #504, opened 2017-05-15
Sk sudo mode

ShaneKilkelly:


app/coffee/router.coffee
```diff
@@ -239,6 +247,9 @@ module.exports = class Router
 		webRouter.get "/beta/participate",  AuthenticationController.requireLogin(), BetaProgramController.optInPage
 		webRouter.post "/beta/opt-in", AuthenticationController.requireLogin(), BetaProgramController.optIn
 		webRouter.post "/beta/opt-out", AuthenticationController.requireLogin(), BetaProgramController.optOut
+		webRouter.get "/confirm-password", AuthenticationController.requireLogin(), SudoModeController.sudoModePrompt
+		webRouter.post "/confirm-password/submit", AuthenticationController.requireLogin(), SudoModeController.submitPassword
```
jpallen: This would be more RESTful as just a `POST /confirm-password`

ShaneKilkelly: True, will change

PR #516, opened 2017-05-24
Sk labels autocomplete

ShaneKilkelly:
Client-side label autocpmplete


public/coffee/ide/labels/LabelsManager.coffee
```diff
@@ -0,0 +1,59 @@
+define [
```
briangough: One option is for this to go under public/coffee/ide/editor/directives/aceEditor like the spell check and other editor extensions.

public/coffee/ide/labels/LabelsManager.coffee
```diff
@@ -0,0 +1,59 @@
...
+						() =>
+							@periodicLoadInterval = setInterval(
+								() =>
+									@loadLabelsFromOpenDoc()
+								, AUTOMATIC_REFRESH_PERIOD
+							)
+						, AUTOMATIC_REFRESH_PERIOD
+					)
+					# listen for document open
+					@$scope.$on 'document:opened', (e, doc) =>
```
briangough: there should be an ace event for the new document, to avoid needing a timeout of 1second if that is why it is there?

public/coffee/ide/labels/LabelsManager.coffee
```diff
@@ -0,0 +1,59 @@
...
+		constructor: (@ide, @$scope) ->
+			@$scope.$root._labels = this
+
+			@state =
+				documents: {} # map of DocId => List[Label]
+
+			@loadLabelsTimeout = null
+			@periodicLoadInterval = null
+
+			setTimeout(
```
briangough: I think its a good idea to avoid these background processes if possible.  We should be able to listen on changes and process the corresponding doc after a delay.  That would catch everything and also work for documents being updated by other users if the event handler is on the document.

ShaneKilkelly:
> I think this can work without needing so much polling by listening to document change events.

The change-listening is done in `AutocompleteManager`, see: https://github.com/sharelatex/web-sharelatex/pull/516/files#diff-dbda54d5bbdae81c8eb156b28c94def8R134

I think you're right though, in that the change listener should cover all cases, and so the periodic scan is not required.

public/coffee/ide/editor/directives/aceEditor/auto-complete/AutoCompleteManager.coffee
```diff
@@ -89,18 +121,33 @@ define [
 		onChange: (change) ->
```
briangough: can you move this labels specific code into LabelsManager so that module is self contained
I think it can work just like the SpellCheckManager.

ShaneKilkelly: Ah, I see. That's a much better way to handle it, will refactor.

PR #518, opened 2017-06-01
Aggregate adjacent inserts and deletes

paulojreis:
Changes:

* The `updateEntries` cycle now includes logic to determine if changes should be aggregated (i.e. a deletion just next to an insertion, made by the same user);
* A new directive to show aggregated change entries;
* The logic to determine selected changes (for bulk actions) is different:
    * `$scope.reviewPanel.selectedEntryIds` contains all selected changes—if an aggregate change is selected, then two ids will be present (the insertion + the deletion);
    * `$scope.reviewPanel.nVisibleSelectedChanges` is the number of entries from a user PoV. An aggregate change counts as one (although it contains two changes). This is used to be shown in the UI and to determine bulk actions visibility.
* Rejection of multiple changes is now done in reverse order.


jpallen:
Approach looks sensible and clean. A few points:

1) 'Agg' is not a clear acronym. Can we change e.g. `agg-change` -> `aggregate-change`, `acceptAggChange` -> `acceptAggregateChange` to be more explicit.
2) Instead of having additional `agg_op` and `agg_op_id` properties, can we instead treat both parts of the change symmetrically. I.e. An array of ops and op_ids that contain two elements.
3) This might let us simplify all the code, and avoid the need for a separate `acceptAggChange` methods. If all entries contained an array of their ops and ids (with inserts and deletes only having one entry, and aggregate changes having two), then one code path could deal with them all. It's a little more refactoring, but I think would be cleaner in total, since we'd only need to understand one model/data structure.

All of this is just from a cursory look, so we should probably discuss in more detail before deciding anything.

app/views/project/editor/review-panel.pug
```diff
@@ -147,6 +158,17 @@
 							permissions="permissions"
 						)
 
+					div(ng-if="entry.type === 'agg-change'")
+						aggregate-change-entry(
+							entry="entry"
+							user="users[entry.metadata.user_id]"
+							on-reject="rejectChange(entry_id);"
+							on-accept="acceptChange(entry_id);"
```
jpallen: should these call the `[accept|reject]AggChange` methods?

PR #522, opened 2017-06-07
set text layer to display:none for faster scroll

briangough:
this needs some work, as we would like to select text across multiple pages -- as is currently possible,
but the scrolling is MUCH faster


paulojreis:
@briangough just added a commit to address that issue—we'll now only consider clicks within the PDF pages. Clicking on the grey bg area or on the scrollbar area won't turn the text layer on.

briangough:
:+1:

PR #525, opened 2017-06-09
Sk server side labels loading

ShaneKilkelly:


app/coffee/Features/Labels/LabelsController.coffee
```diff
@@ -0,0 +1,26 @@
+EditorRealTimeController = require "../Editor/EditorRealTimeController"
+LabelsHandler = require './LabelsHandler'
+logger = require 'logger-sharelatex'
+
+
+module.exports = LabelsController =
+
+	getAllLabels: (req, res, next) ->
+		project_id = req.params.Project_id
```
jpallen: We should default to lower case `project_id` for new methods. We'll get everything else migrated eventually!

app/coffee/Features/Labels/LabelsController.coffee
```diff
@@ -0,0 +1,26 @@
...
+		LabelsHandler.getAllLabelsForProject project_id, (err, projectLabels) ->
+			if err?
+				logger.err {project_id, err}, "[LabelsController] error getting all labels from project"
+				return next(err)
+			res.json {projectId: project_id, projectLabels: projectLabels}
+
+	getLabelsForDoc: (req, res, next) ->
+		project_id = req.params.Project_id
```
jpallen: As above

app/coffee/Features/Labels/LabelsHandler.coffee
```diff
@@ -0,0 +1,50 @@
...
+
+	extractLabelsFromProjectDocs: (docs, callback=(err, projectLabels)->) ->
+		projectLabels = {}  # docId => List[Label]
+		for _docPath, doc of docs
+			docLabels = []
+			for line in doc.lines
+				re = LabelsHandler.labelCaptureRegex()
+				while (labelMatch = re.exec(line))
+					if labelMatch[1]
+						docLabels.push(labelMatch[1])
```
jpallen: This could call `extractLabelsFromDoc` to be more DRY?

app/coffee/Features/Labels/LabelsController.coffee
```diff
@@ -0,0 +1,26 @@
...
+		doc_id = req.params.doc_id
+		LabelsHandler.getLabelsForDoc project_id, doc_id, (err, docLabels) ->
+			if err?
+				logger.err {project_id, doc_id, err}, "[LabelsController] error getting labels from doc"
+				return next(err)
+			EditorRealTimeController.emitToRoom project_id, 'docLabelsUpdated', {
```
jpallen: This is surprising to find here, given that we haven't updated anything - can you add a comment to explain why this is called here.

app/coffee/Features/Labels/LabelsHandler.coffee
```diff
@@ -0,0 +1,50 @@
...
+		ProjectEntityHandler.getAllDocs projectId, (err, docs) ->
+			if err?
+				return callback(err)
+			LabelsHandler.extractLabelsFromProjectDocs docs, (err, projectLabels) ->
+				if err?
+					return callback(err)
+				callback(null, projectLabels)
+
+	getLabelsForDoc: (projectId, docId, callback=(err, docLabels)->) ->
+		DocumentUpdaterHandler.flushDocToMongo projectId, docId, (err) ->
```
jpallen: `getLabelsForDoc` flushes first, but `getAllLabelsForProject` doesn't. Is this correct, if so can you add a comment to explain why.

app/views/project/editor/editor.pug
```diff
@@ -37,6 +37,7 @@ div.full-size(
 			keybindings="settings.mode",
 			font-size="settings.fontSize",
 			auto-complete="settings.autoComplete",
+			enable-auto-complete-features="true",
```
jpallen: We could only enable this only for users with permission to write, and likewise with the permissions on the server side routes. Not really essential, but reduce load slightly, and particularly in the case of a large number of read-only views of a project if it was public say.

app/views/project/editor/editor.pug
```diff
@@ -60,6 +61,7 @@ div.full-size(
 			track-changes= "editor.trackChanges",
 			doc-id="editor.open_doc_id"
 			renderer-data="reviewPanel.rendererData"
+			labels-master="labelsMaster"
```
jpallen: Is this unused now?

public/coffee/ide.coffee
```diff
@@ -127,6 +127,13 @@ define [
 		ide.permissionsManager = new PermissionsManager(ide, $scope)
 		ide.binaryFilesManager = new BinaryFilesManager(ide, $scope)
 
+		# Set up labels
+		$scope.$on 'doc:labels:updated', labels.onDocLabelsUpdated
+		$scope.$on 'entity:deleted', labels.onEntityDeleted
+		$scope.$on 'file:upload:complete', labels.fileUploadComplete
+		$timeout () ->
+			labels.loadProjectLabelsFromServer()
+
```
jpallen: Could these be handled in `LabelsManager` to keep all label related code in one place? Do we need to `rootScope`? That should still be possible in `LabelsManager`

public/coffee/ide/editor/directives/aceEditor.coffee
```diff
@@ -32,11 +33,13 @@ define [
...
 			return url
 
-	App.directive "aceEditor", ($timeout, $compile, $rootScope, event_tracking, localStorage, $cacheFactory) ->
+	App.directive "aceEditor", ($timeout, $compile, $rootScope, event_tracking, localStorage, $cacheFactory, labels) ->
 		monkeyPatchSearch($rootScope, $compile)
+		
 
 		return  {
 			scope: {
+				labelsMaster: '='
```
jpallen: Redundant?

public/coffee/ide.coffee
```diff
@@ -172,5 +179,7 @@ define [
 		# User can append ?ft=somefeature to url to activate a feature toggle
 		ide.featureToggle = location?.search?.match(/^\?ft=(\w+)$/)?[1]
 
+		ide.socket.on 'docLabelsUpdated', (data) ->
+			$scope.$broadcast 'doc:labels:updated', data
```
jpallen: Also handle in `LabelsManager` if possible.

app/coffee/Features/Labels/LabelsController.coffee
jpallen: Would a more semantic name be `broadcastLabelsForDoc` and make it POST request since it has side effects? Does the client that calls this actually need the returned data, or can it get that client side?

public/coffee/ide/editor/directives/aceEditor/labels/LabelsManager.coffee
```diff
@@ -28,37 +26,26 @@ define [
 					((_.any(change.lines, (line) -> line.match(/\\label\{[^\}\n\\]{0,80}\}/))) or
 					 (commandFragment?.length > 2 and commandFragment.slice(0,7) == '\\label{'))
 				)
```
jpallen: It took me a while to parse and figure out what this was doing. Could change it to something like:

```
return if change.action not in ['remove', 'insert']
changeUpdatedFullLabel = (_.any(change.lines, (line) -> line.match(/\\label\{[^\}\n\\]{0,80}\}/)))
isTypingInLabel =  (commandFragment?.length > 2 and commandFragment.slice(0,7) == '\\label{')
if (changeUpdatedFullLabel or isTypingInLabel)
    ...
```

public/coffee/ide.coffee
ShaneKilkelly: Because `LabelsManager` is part of the Ace directive, it get's initialized more than once per page load. It feels cleaner to have the ide-global behaviours be init'd in the ide.

app/coffee/Features/Labels/LabelsController.coffee
ShaneKilkelly: Yeah, I suppose `updated` has the wrong connotations, `broadcast` makes more sense.

public/coffee/ide.coffee
ShaneKilkelly: Hmm, I've tried to move this code, but there seems to be an issue with angular initialisation order, whereby the `ide` isn't fully populated when the factory functions are evaluated. Will park it for now but keep an eye on it with future work.

app/coffee/router.coffee
```diff
@@ -193,6 +194,9 @@ module.exports = class Router
 		webRouter.get  '/Project/:Project_id/download/zip', AuthorizationMiddlewear.ensureUserCanReadProject, ProjectDownloadsController.downloadProject
 		webRouter.get  '/project/download/zip', AuthorizationMiddlewear.ensureUserCanReadMultipleProjects, ProjectDownloadsController.downloadMultipleProjects
 
+		webRouter.get '/project/:project_id/labels', AuthorizationMiddlewear.ensureUserCanReadProject, AuthenticationController.requireLogin(), LabelsController.getAllLabels
+		webRouter.get '/project/:project_id/:doc_id/labels', AuthorizationMiddlewear.ensureUserCanReadProject, AuthenticationController.requireLogin(), LabelsController.getLabelsForDoc
```
jpallen: The doc_id should be namespaced under /doc, i.e. `/project/:project_id/doc/:doc_id/labels`

public/coffee/ide.coffee
jpallen: How about we move this into a `LabelsManager`, like we have with the few lines above, each part of the editor is in it's own little wrapper. So have a folder structure like:

```
coffee/ide/labels/
    LabelsManager.coffee
    services/
        labels.coffee
```

to keep it all in one place.

public/coffee/ide/editor/directives/aceEditor.coffee
```diff
@@ -10,6 +10,7 @@ define [
 	"ide/editor/directives/aceEditor/cursor-position/CursorPositionManager"
 	"ide/editor/directives/aceEditor/track-changes/TrackChangesManager"
 	"ide/editor/directives/aceEditor/labels/LabelsManager"
+	"services/labels"
```
jpallen: Namespace this under `ide/labels/services/labels` as above.

app/coffee/Features/Labels/LabelsController.coffee
jpallen: Sorry, I think what I meant wasn't clear here. I think it's the parent `getLabelsForDoc` that is not semantically named, and it's odd that we're doing some broadcast of something when all we're doing is GETing. I think the top level HTTP method should be something like `broadcastLabelsForDoc` and should be a post request. The client doesn't actually care about the HTTP response anyway right, because it gets updated through socket io, so it's not even using the GET response at the moment?

app/coffee/Features/Labels/LabelsController.coffee
```diff
@@ -0,0 +1,26 @@
...
+
+	getAllLabels: (req, res, next) ->
+		project_id = req.params.project_id
+		LabelsHandler.getAllLabelsForProject project_id, (err, projectLabels) ->
+			if err?
+				logger.err {project_id, err}, "[LabelsController] error getting all labels from project"
+				return next(err)
+			res.json {projectId: project_id, projectLabels: projectLabels}
+
+	getLabelsForDoc: (req, res, next) ->
```
jpallen: Pasting from other comment to make this part of the review: Sorry, I think what I meant wasn't clear here. I think it's the parent `getLabelsForDoc` that is not semantically named, and it's odd that we're doing some broadcast of something when all we're doing is GETing. I think the top level HTTP method should be something like broadcastLabelsForDoc and should be a post request. The client doesn't actually care about the HTTP response anyway right, because it gets updated through socket io, so it's not even using the GET response at the moment?

app/coffee/Features/Labels/LabelsHandler.coffee
```diff
@@ -0,0 +1,56 @@
...
+				return callback(err)
+			LabelsHandler.extractLabelsFromProjectDocs docs, (err, projectLabels) ->
+				if err?
+					return callback(err)
+				callback(null, projectLabels)
+
+	getLabelsForDoc: (projectId, docId, callback=(err, docLabels)->) ->
+		# Flush doc first, because this action is often performed while
+		# a document is being edited by the client. By contrast,
+		# `getAllLabelsForProject` is called only when a project/editor is loaded
```
jpallen: I think we should actually flush in both cases, i.e. also in `getAllLabelsForProject`

app/coffee/Features/Labels/LabelsHandler.coffee
```diff
@@ -0,0 +1,56 @@
...
+		callback(null, docLabels)
+
+	extractLabelsFromProjectDocs: (projectDocs, callback=(err, projectLabels)->) ->
+		projectLabels = {}  # docId => List[Label]
+		docs = for _path, doc of projectDocs
+			doc
+		Async.eachSeries(
+			docs
+			, (doc, cb) ->
+				LabelsHandler.extractLabelsFromDoc doc.lines, (err, docLabels) ->
```
jpallen: async can go a bit screwy if you're not actually doing asynchronous stuff in it. If we get rid of the callback pattern in these methods, since they are synchronous, this would be a simple for loop and not have the async issue.

ShaneKilkelly:
@jpallen done :)

app/views/project/editor/editor.pug
```diff
@@ -37,9 +37,9 @@ div.full-size(
 			keybindings="settings.mode",
 			font-size="settings.fontSize",
 			auto-complete="settings.autoComplete",
-			enable-auto-complete-features="true",
+			enable-auto-complete-features="!anonymous"
```
jpallen: Anonymous users may have write permissions, so this should be `permissions.write` (see the read-only attribute on the editor below on line 51).

jpallen:
@ShaneKilkelly One more small comment, but it's simple. Happy for this to roll out after that. This is really a nice feature that's come together quickly, good job! :)

PR #529, opened 2017-06-16
Spellcheck: ignore args to some latex commands

ShaneKilkelly:


public/coffee/ide/editor/directives/aceEditor/spell-check/SpellCheckManager.coffee
```diff
@@ -53,6 +53,9 @@ define [
 			@$scope.learnWord = (highlight) =>
 				@learnWord(highlight)
 
+			# DEBUG
```
jpallen: 🙃

public/coffee/ide/editor/directives/aceEditor/spell-check/SpellCheckManager.coffee
```diff
@@ -246,3 +250,28 @@ define [
...
 			return $.ajax options
+
+		blacklistedCommandRegex: ///
+			\\                    # initial backslash
+			(label                # any of these commands
+			|[a-z]{0,2}ref
+			|usepackage
+			|begin
+			|end
+			|[a-z]{0,2}cite
```
jpallen: There are `\...cite` commands with more than two prefix letters, e.g. `\parencite`. I'd imagine similarly for `\ref`. Is the concern about performance/scanning far ahead? If so, just something big like 16?

public/coffee/ide/editor/directives/aceEditor/spell-check/SpellCheckManager.coffee
```diff
@@ -246,3 +250,28 @@ define [
...
+			|input
+			|include
+			|includegraphics)
+			( \[ [^\]]* \] )?     # optional [...] args
+				\{ [^}]*  \}        # the {...} args
+		///g
+
+		blankOutBlacklistedCommands: (line) ->
+			line.replace @blacklistedCommandRegex, (command) ->
+				command.replace(
```
jpallen: Why do we need to do any analysis of the match, can we not just blank the whole thing? Am I right in saying this currently returns, e.g. `\usepackage[.........]{.............}`. Can we just return all `.............................` instead?

ShaneKilkelly: Correct, and yes, the latter makes more sense :)

PR #530, opened 2017-06-19
Add ability to show invite link in Share modal

ShaneKilkelly:


natestemen:
pretty sure this was around the time of the sharelatex/overleaf merger. probably left over from that. okay to close?

PR #532, opened 2017-06-21
Support per user track changes (proof of concept)

jpallen:


paulojreis:
This is ready to be reviewed.

Before putting it live, we need to change https://github.com/sharelatex/web-sharelatex/pull/532/files#diff-7701684cedd84223921f73f784273c4eR239 to the specific date of the deploy.

PR #538, opened 2017-06-30
Sk ac ignore comments

ShaneKilkelly:


public/coffee/ide/editor/directives/aceEditor/auto-complete/AutoCompleteManager.coffee
```diff
@@ -121,6 +121,8 @@ define [
 			end = change.end
 			range = new Range(end.row, 0, end.row, end.column)
 			lineUpToCursor = @editor.getSession().getTextRange(range)
+			if lineUpToCursor.slice(0, 1) == '%'
```
jpallen: This will only catch comments like:

```
% This is a full line comment
```

Not comments like:

```
$x^2$ % This is  partial line comment.
```

Could just make it a regex looking for any % (Without a preceding \) before the cursor.

public/coffee/ide/editor/directives/aceEditor/auto-complete/SuggestionManager.coffee
```diff
@@ -39,15 +39,18 @@ define [], () ->
 			return commands
 
 		# Ignore single letter commands since auto complete is moot then.
-		commandRegex: /\\([a-zA-Z][a-zA-Z]+)/
+		# Ignore commands which occur after a comment character '%'
+		commandRegex: /^([^%]*?)\\([a-zA-Z][a-zA-Z]+)/m
```
jpallen: Need to ignore `\%` characters still though.

jpallen: I also worry this regexp could be expensive now since it has to backtrack? How about blanking all comments first, then running the current regex? That might be easier... not sure.

public/coffee/ide/editor/directives/aceEditor/auto-complete/AutoCompleteManager.coffee
ShaneKilkelly: Good catch, will fix

public/coffee/ide/editor/directives/aceEditor/auto-complete/SuggestionManager.coffee
ShaneKilkelly: Blanking will probably be cheaper, yes. Will try a solution with blanking.

PR #539, opened 2017-06-30
Don't trigger autocomplete when editing existing environment

ShaneKilkelly:


public/coffee/ide/editor/directives/aceEditor/auto-complete/AutoCompleteManager.coffee
```diff
@@ -122,6 +125,9 @@ define [
 			range = new Range(end.row, 0, end.row, end.column)
 			lineUpToCursor = @editor.getSession().getTextRange(range)
 			commandFragment = getLastCommandFragment(lineUpToCursor)
+			commandName = getCommandNameFromFragment(commandFragment)
+			if commandName in ['begin', 'end']
```
jpallen: This looks like it would stop the auto-complete running on \begin{| (i.e. creating an environment for the first time. I think we only want to not auto-complete if the \begin{foo} is already complete (i.e. has a closing brace perhaps - although this might not work as a test if we add in auto-closing braces, perhaps a closing brace AND some existing environment name?)

ShaneKilkelly: Nope, autocomplete triggers correctly when typing out the initial `\begin`, will put a gif in the ticket to demo

ShaneKilkelly: But the test could be made tighter still

PR #541, opened 2017-07-05
add public api router

briangough:


app/coffee/infrastructure/Modules.coffee
```diff
@@ -15,14 +15,14 @@ module.exports = Modules =
 				@modules.push loadedModule
 		Modules.attachHooks()
 
-	applyRouter: (webRouter, apiRouter) ->
+	applyRouter: (webRouter, apiRouter, publicApiRouter) ->
```
henryoswald: shall we rename apiRouter -> privateApiRouter to be explicit?

PR #543, opened 2017-07-06
Enable (optional) auto-pairing for latex mode

ShaneKilkelly:


jpallen:
Doesn't work when there are autocomplete options shown:

PR #555, opened 2017-07-24
Enforce stricter password policy.

ShaneKilkelly:
- Check minimum password lengths
- Set default policy to 6-128 chars


jpallen:
Am I right in saying that we only do complex validation on the client side? On the server we only check min and max length (which is all we're using on sharelatex.com anyway?)

ShaneKilkelly:
I think so, yes, but even on the client it doesn't seem to do much more (at least that's noticable) beyond length validation.

jpallen:
Yea, I think that's just an option that can be enabled for onsite installs

PR #561, opened 2017-07-31
Persist `autoPairDelimiters` setting.

ShaneKilkelly:


jdleesmiller:
On background... is there any validation of `req.body.autoPairDelimiters` before it gets stored? We've had problems over the years with bots injecting junk into form fields, which when not caught by validation caused some strange issues downstream.

jpallen:
No, I don't think there is. An Mongo won't even cast a string to a boolean, etc. One quick fix (for booleans) in JS is `!!req.body. autoPairDelimiters` to force it into a boolean. String based settings would be harder, and probably need a whitelist, or at least a check it is a string.

jpallen:
Ah, although this particular code goes through Mongoose (an ORM), which uses the schema to cast things to the right type. So in this case, it's probably fine. We're generally moving away from Mongoose due to the overhead though, and just using mongojs, which is more direct mongo driver in the rest of the code base.

jdleesmiller:
Aha, I see --- now I've found [the schema](https://github.com/sharelatex/web-sharelatex/blob/c35c00200a9a80b61d1b9d0b9d1d14eb1ae1646f/app/coffee/models/User.coffee), which is useful.

For my info: is that runtime overhead, or some other type of overhead?

jpallen:
Yes, runtime overhead. Mongoose does a lot more work at the parsing / serializing steps of talking to Mongo. Particularly on the large nested Project schema, it can be an order of magnitude slower than just parsing/serializing the Mongo BSON response to/from a plain JS object, which is what mongojs does.

jdleesmiller:
OK, thanks for the explanation!

PR #564, opened 2017-07-31
Sk autocomplete graphics

ShaneKilkelly:


public/stylesheets/app/editor.less
```diff
@@ -504,3 +504,8 @@
 	height: auto;
 	border-bottom: 1px solid @modal-header-border-color;
 }
+
+// Widen autocomplete popup
+.ace_autocomplete {
+	width: 380px !important;
```
jpallen: I see the problem that this is trying to solve, but I don't think this is the right way to do it. Even with this, it doesn't take much of a file name before it doesn't fit, and this makes it oddly wide for the rest of the auto-complete behaviour. Is there any possibility of dynamically resizing the pop-up to the content size?

jpallen: I see the problem that this is trying to solve, but I don't think this is the right way to do it. Even with this, it doesn't take much of a file name before it doesn't fit, and this makes it oddly wide for the rest of the auto-complete behaviour. Is there any possibility of dynamically resizing the pop-up to the content size?

public/coffee/ide/preamble/preamble/services/preamble.coffee
```diff
@@ -0,0 +1,14 @@
...
+	App.factory 'preamble', (ide) ->
+
+		Preamble = {
+			getPreambleText: () ->
+				text = ide.editorManager.getCurrentDocValue().slice(0, 5000)
+				preamble = text.match(/([^]*)^\\begin\{document\}/m)[1]
+				
+		}
+
+		return Preamble
```
jpallen: Duplicate dead file?

public/stylesheets/app/editor.less
ShaneKilkelly: I had created another ticket yesterday to look into this, but having had a chance to ponder the issue it might be reasonably straight-forward, will try it today.

public/coffee/ide/preamble/preamble/services/preamble.coffee
ShaneKilkelly: Ah yes, will delete.

public/coffee/ide/editor/directives/aceEditor/auto-complete/AutoCompleteManager.coffee
```diff
@@ -245,7 +282,18 @@ define [
...
 						editor.completer.showPopup(editor)
 						editor.completer.cancelContextMenu()
-						$(editor.completer.popup?.container).css({'font-size': @$scope.fontSize + 'px'})
+						container = $(editor.completer.popup?.container)
+						container.css({'font-size': @$scope.fontSize + 'px'})
+						# Dynamically set width of autocomplete popup
+						if filtered = editor?.completer?.completions?.filtered
+							longestCaption = _.max(filtered.map( (c) -> c.caption.length ))
+							longestMeta = _.max(filtered.map( (c) -> c.meta.length ))
+							charScale = @$scope.fontSize * 0.7
```
jpallen: This can be got directly via `editor.renderer.characterWidth`

ShaneKilkelly: Ah, good call

PR #565, opened 2017-08-01
Ho overleaf styling based on domain name

henryoswald:


config/settings.defaults.coffee
```diff
@@ -16,6 +16,15 @@ httpAuthUsers[httpAuthUser] = httpAuthPass
...
 module.exports = settings =
+
+
+	# Overleaf overide settings
+	# -------------------------
+	# These should be named the same as corisponding ShareLaTeX setting when possible.
+	overleaf: 
+
+		cookieDomain:".overleaf.local"
+
```
henryoswald: we could have a separate overleaf config file still with overides that are read when the site is in overleaf mode

PR #568, opened 2017-08-02
Validate and add error status to contact form

jpallen:
If we send a ticket to groove without either a subject, email or body, it rejects it. However, we had no error handling on the response from Groove, so this would be silently ignored, and the user told that their request had been submitted.


jpallen:
Note that the only changes to the form section are indenting it to add a `form` tag around it, so that we can add the angular validation properties to the subject, email and body

PR #569, opened 2017-08-02
compile from redis

briangough:
keep project state in redis, try to compile incrementally when possible (state hasn't changed), otherwise retry with a full compile

Outstanding tasks:
- [x] make project state hash deterministic
- [x] add tests

### DIFF
--- a/app/coffee/Features/Compile/ClsiManager.coffee
+++ b/app/coffee/Features/Compile/ClsiManager.coffee
@@ -25,7 +25,7 @@ module.exports = ClsiManager =
 				callback(error, status, result...)
 
 	sendRequestOnce: (project_id, user_id, options = {}, callback = (error, status, outputFiles, clsiServerId, validationProblems) ->) ->
-		ClsiManager._buildRequest project_id, user_id, options, (error, req) ->
+		ClsiManager._buildRequest project_id, options, (error, req) ->
 			return callback(error) if error?
 			logger.log project_id: project_id, "sending compile to CLSI"
 			console.log "REQUEST", JSON.stringify(req, null, 2)
@@ -119,7 +119,7 @@ module.exports = ClsiManager =
 
 	VALID_COMPILERS: ["pdflatex", "latex", "xelatex", "lualatex"]
 
-	_buildRequest: (project_id, user_id, options={}, callback = (error, request) ->) ->
+	_buildRequest: (project_id, options={}, callback = (error, request) ->) ->
 		Project.findById project_id, {}, (error, project) ->
 			return callback(error) if error?
 			return callback(new Errors.NotFoundError("project does not exist: #{project_id}")) if !project?
@@ -127,7 +127,7 @@ module.exports = ClsiManager =
 			if project.compiler not in ClsiManager.VALID_COMPILERS
 				project.compiler = "pdflatex"
 
-			ClsiStateManager.checkState project_id, user_id, project, (error, stateOk, state) ->
+			ClsiStateManager.checkState project_id, project, (error, stateOk, state) ->
 				return callback(error) if error?
 				logger.log project_id: project_id, checkState: stateOk, "checked project state"
 				console.log "OPTIONS ARE", options
@@ -151,7 +151,7 @@ module.exports = ClsiManager =
 						return callback(error) if error?
 						console.log "DOCS", docs
 						# FIXME want to store state after project has been sent to clsi
-						ClsiStateManager.setState project_id, user_id, project, (error, state) ->
+						ClsiStateManager.setState project_id, project, (error, state) ->
 							if error?
 								logger.err err:error, project_id:project_id, "error storing state in redis"
 								#return callback(error)
--- a/app/coffee/Features/Compile/ClsiManager.coffee
+++ b/app/coffee/Features/Compile/ClsiManager.coffee
@@ -10,6 +10,7 @@ ClsiCookieManager = require("./ClsiCookieManager")
 _ = require("underscore")
 async = require("async")
 ClsiFormatChecker = require("./ClsiFormatChecker")
+DocumentUpdaterHandler = require "../DocumentUpdater/DocumentUpdaterHandler"
 
 module.exports = ClsiManager =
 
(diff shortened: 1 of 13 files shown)


app/coffee/Features/Compile/ClsiManager.coffee
```diff
@@ -7,16 +7,28 @@ ProjectEntityHandler = require("../Project/ProjectEntityHandler")
...
+DocumentUpdaterHandler = require "../DocumentUpdater/DocumentUpdaterHandler"
 
 module.exports = ClsiManager =
 
-	sendRequest: (project_id, user_id, options = {}, callback = (error, status, outputFiles, clsiServerId, validationProblems) ->) ->
+	sendRequest: (project_id, user_id, options = {}, callback) ->
+		ClsiManager.sendRequestOnce project_id, user_id, _.clone(options), (error, status, result...) ->
+			return callback(error) if error?
+			if status is 'conflict'
+				options.state = "conflict" # will force full compile
```
jpallen: Why do we set the state to `conflict` here rather than setting `incremental` to `false`?

app/coffee/Features/Compile/ClsiManager.coffee
```diff
@@ -101,56 +118,101 @@ module.exports = ClsiManager =
 		return outputFiles
 
 	VALID_COMPILERS: ["pdflatex", "latex", "xelatex", "lualatex"]
+
 	_buildRequest: (project_id, options={}, callback = (error, request) ->) ->
-		Project.findById project_id, {compiler: 1, rootDoc_id: 1, imageName: 1}, (error, project) ->
+		Project.findById project_id, {}, (error, project) ->
```
jpallen: We may want to look at doing this with mongojs rather than mongoose to avoid the ORM overhead.

jpallen: We should also still filter to just the previous attributes and `rootFolder`.

app/coffee/Features/Compile/ClsiManager.coffee
```diff
@@ -101,56 +118,101 @@ module.exports = ClsiManager =
...
 
+			ClsiStateManager.checkState project_id, project, (error, stateOk, state) ->
+				return callback(error) if error?
+				logger.log project_id: project_id, checkState: stateOk, "checked project state"
+				console.log "OPTIONS ARE", options
+				if stateOk and not options.state? # incremental
+					ClsiManager._getContentFromDocUpdater project_id, (error, docUpdaterDocs) ->
+						return callback(error) if error?
+						# make this incremental
+						ProjectEntityHandler.getAllDocPathsFromProject project, (error, docPath) ->
```
jpallen: We don't seem to load the files here. Is that a WIP?

app/coffee/Features/Project/ProjectEntityHandler.coffee
```diff
@@ -74,6 +73,29 @@ module.exports = ProjectEntityHandler =
 						files[path.join(folderPath, file.name)] = file
 			callback null, files
 
+	getAllFoldersFromProject: (project, callback) ->
+		folders = {}
+		processFolder = (basePath, folder) ->
```
jpallen: This method is still included in `getAllFolders` above, but is now dead?

jpallen:
As a general comment, I'm struggling to understand what the `state` is just from reading the code. Can we rename this to something more descriptive like `syncState`, and also enumerate the possible states it can be in explicitly in the code somewhere (perhaps an ENUM construction somewhere?)

Likewise, what does `checkState` being `ok` mean? Would it be reasonable to call that `checkSynced`?

jpallen:
I've added a lot of small comments, but overall this approach looks clean and clear 👍

app/coffee/Features/Compile/ClsiManager.coffee
briangough: Long story: I started that way, but it was slightly ugly as I had to delete it from the options on the full request, because if it got passed through the the CLSI the RequestParser there didn't like it because false wasn't a string.

It's a sign the current method isn't the right one  I think, maybe the following might work better:

compile_type: "full" | "incremental"
state: "hash value"

app/coffee/Features/Compile/ClsiManager.coffee
briangough: For an incremental compile we shouldn't need the files?  The files on disk in the clsi won't be removed.

app/coffee/Features/Compile/ClsiManager.coffee
briangough: Done, and switched to mongojs

app/coffee/Features/Project/ProjectEntityHandler.coffee
briangough: Yes, good catch.

briangough:
I've changed checkState to checkProjectStateMatch with result in projectStateUnchanged, and replaced state by projectStateHash

app/coffee/Features/Compile/ClsiManager.coffee
```diff
@@ -101,56 +117,116 @@ module.exports = ClsiManager =
...
 	_buildRequest: (project_id, options={}, callback = (error, request) ->) ->
-		Project.findById project_id, {compiler: 1, rootDoc_id: 1, imageName: 1}, (error, project) ->
+		ProjectGetter.getProject project_id, {compiler: 1, rootDoc_id: 1, imageName: 1, rootFolder:1}, (error, project) ->
 			return callback(error) if error?
 			return callback(new Errors.NotFoundError("project does not exist: #{project_id}")) if !project?
-
 			if project.compiler not in ClsiManager.VALID_COMPILERS
 				project.compiler = "pdflatex"
 
+			if options.incremental or options.syncType? # new way, either incremental or full
```
jpallen: this `options.incremental` is a feature flag I think? It reads as if it will make it an `incremental` compile at the moment, so perhaps call this `options.incrementalCompilesEnabled`?

app/coffee/Features/Compile/ClsiStateManager.coffee
```diff
@@ -0,0 +1,27 @@
...
+#
+# When the hash changes the full set of files on the CLSI will need to
+# be updated.  If it doesn't change then we can overwrite changed docs
+# in place on the clsi, getting them from the docupdater.
+#
+# The docupdater is responsible for setting the key in redis, and
+# unsetting it if it removes any documents from the doc updater.
+
+buildState = (project) ->
+	json = JSON.stringify(project.rootFolder)
```
jpallen: JSON.stringify isn't guaranteed to produce the same string each time from the same object. It's generally reliable with small objects, but we shouldn't rely on it, and the keys of an object can be in any order. I think we'll need to build up our own string representation to be safe.

app/coffee/Features/DocumentUpdater/DocumentUpdaterHandler.coffee
```diff
@@ -123,6 +123,30 @@ module.exports = DocumentUpdaterHandler =
...
+	getProjectDocsIfMatch: (project_id, projectStateHash, callback = (error, docs) ->) ->
+		timer = new metrics.Timer("get-project-docs")
+		url = "#{settings.apis.documentupdater.url}/project/#{project_id}/doc?state=#{projectStateHash}"
+		logger.log project_id:project_id, "getting project docs from document updater"
+		request.get url, (error, res, body)->
+			timer.done()
+			if error?
+				logger.error err:error, url:url, project_id:project_id, "error getting project docs from doc updater"
+				return callback(error)
+			if res.statusCode is 409
```
jpallen: Can you add a comment about why this code is special and what it means

app/coffee/Features/Compile/ClsiStateManager.coffee
briangough: I can change it to be a list of sorted list file paths+ids.  I figured it would be deterministic retrieving the same object from mongo in the same version of node but maybe not worth taking a chance.

test/UnitTests/coffee/Compile/ClsiManagerTests.coffee
```diff
@@ -103,6 +103,25 @@ describe "ClsiManager", ->
...
+				@ClsiManager.sendRequest @project_id, @user_id, {}, @callback
+
+			it "should call the sendRequestOnce method twice", ->
+				@ClsiManager.sendRequestOnce.calledTwice.should.equal true
+
+			it "should call the sendRequestOnce method once with syncType:full", ->
+				@ClsiManager.sendRequestOnce.withArgs(@project_id, @user_id, {syncType:"full"}).calledOnce.should.equal true
+
+			it "should call the sendRequestOnce method once without syncType:full", ->
+				@ClsiManager.sendRequestOnce.withArgs(@project_id, @user_id, {}).calledOnce.should.equal true
```
jpallen: I could be wrong, but I think `withArgs` is used for setting up the stub, and you need to use `calledWithArgs` for the test. I think all this is checking is the `calledOnce` criteria?

jpallen:
One comment on 6a46cfebf2fdbac0ed75a6965806d36a11997b9c, but otherwise looks good 👍

test/UnitTests/coffee/Compile/ClsiManagerTests.coffee
briangough: I've changed it to be in our usual format with calledWith.  The original works (because a stub is a spy, and the spy api supports the argument match with calledOnce) but it's more confusing so better to check it our normal way.

PR #574, opened 2017-08-07
Add in overleaf styles alongside sharelatex styles

jpallen:
Similar to #565, but based all on settings file config.


jpallen:
Note that the loading logo is not currently filling up with the brand colour as it should.

PR #575, opened 2017-08-08
Allow OAuth based log-ins from Overleaf

jpallen:


jpallen:
The module is here: https://github.com/sharelatex/overleaf-auth-web-module/commit/e94b60c5e2d1eb72e63717f35b995e32f0e75273 It's a single commit rather than PR because it's the first commit...

jpallen:
This should also go out with this: https://github.com/sharelatex/web-sharelatex-modules/pull/15

app/coffee/Features/User/UserCreator.coffee
```diff
@@ -17,23 +17,15 @@ module.exports = UserCreator =
...
-			user.last_name = opts.last_name
-		else
-			user.last_name = ""
+		if !opts.first_name? or opts.first_name == ""
+			opts.first_name = username[0]
 
+		for key, value of opts
+			user[key] = value
+			
+		user.ace.syntaxValidation = true
```
henryoswald: should this be a default on user.coffee?

jpallen: This is just copied from the old code, but I think the fear of a default is it will get set to `true` for users that don't yet have a value (old users) if we do `user.save` somewhere else.

jpallen: The 'proper' solution would be to backfill other users to `false`, but I think that's outside of the scope of this.

PR #584, opened 2017-08-30
Re-enable autocomplete for environments

ShaneKilkelly:


ShaneKilkelly:
Leaving this as it is, because it's a partial fix which opens up another rabbit-hole.

PR #588, opened 2017-09-01
Load user features alongside id and email.

ShaneKilkelly:
Fixes a bug where project features were not applied properly,
and instead fell back to the free-account defaults.


jpallen:
Can you add an acceptance test that would have caught this please (if it's reasonably straightforward)

jpallen:
I think we also need these properties: https://github.com/sharelatex/web-sharelatex/blob/master/app/coffee/Features/Project/ProjectEditorHandler.coffee#L55

ShaneKilkelly:
Good catch

ShaneKilkelly:
@jpallen, added acceptance test

jpallen:
Thanks, good spot and good fix. This could have been ugly!

henryoswald:
we should get the acceptance tests running on the ci box.

PR #591, opened 2017-09-04
Client-side auto compile

40thieves:
### Description

Listens for server saved events and runs auto-compile.

### DIFF
--- a/app/views/project/editor/pdf.pug
+++ b/app/views/project/editor/pdf.pug
@@ -27,14 +27,16 @@ div.full-size.pdf(ng-controller="PdfController")
 				span.caret
 			ul.dropdown-menu.dropdown-menu-left
 				// Only show on beta program?
-				li.dropdown-header Auto-compile
-				li
-					a(href, ng-click="autocompile_enabled = true")
-						i.fa.fa-fw(ng-class="{'fa-check': autocompile_enabled}")
-						| &nbsp;On
-					a(href, ng-click="autocompile_enabled = false")
-						i.fa.fa-fw(ng-class="{'fa-check': !autocompile_enabled}")
-						| &nbsp;Off
+				if user.betaProgram
+					li.dropdown-header Auto-compile
+					li
+						a(href, ng-click="autocompile_enabled = true")
+							i.fa.fa-fw(ng-class="{'fa-check': autocompile_enabled}")
+							| &nbsp;On
+					li
+						a(href, ng-click="autocompile_enabled = false")
+							i.fa.fa-fw(ng-class="{'fa-check': !autocompile_enabled}")
+							| &nbsp;Off
 				li.dropdown-header #{translate("compile_mode")}
 				li
 					a(href, ng-click="draft = false")
--- a/app/views/project/editor/pdf.pug
+++ b/app/views/project/editor/pdf.pug
@@ -28,15 +28,15 @@ div.full-size.pdf(ng-controller="PdfController")
 			ul.dropdown-menu.dropdown-menu-left
 				// Only show on beta program?
 				if user.betaProgram
-					li.dropdown-header Auto-compile
+					li.dropdown-header #{translate("auto_compile")}
 					li
 						a(href, ng-click="autocompile_enabled = true")
 							i.fa.fa-fw(ng-class="{'fa-check': autocompile_enabled}")
-							| &nbsp;On
+							| &nbsp;#{translate('on')}
 					li
 						a(href, ng-click="autocompile_enabled = false")
 							i.fa.fa-fw(ng-class="{'fa-check': !autocompile_enabled}")
-							| &nbsp;Off
+							| &nbsp;#{translate('off')}
 				li.dropdown-header #{translate("compile_mode")}
 				li
 					a(href, ng-click="draft = false")
--- a/app/views/project/editor/pdf.pug
+++ b/app/views/project/editor/pdf.pug
@@ -26,6 +26,15 @@ div.full-size.pdf(ng-controller="PdfController")
 			)
 				span.caret
 			ul.dropdown-menu.dropdown-menu-left
+				// Only show on beta program?
+				li.dropdown-header Auto-compile
+				li
+					a(href, ng-click="autocompile_enabled = true")
+						i.fa.fa-fw(ng-class="{'fa-check': autocompile_enabled}")
+						| &nbsp;On
+					a(href, ng-click="autocompile_enabled = false")
+						i.fa.fa-fw(ng-class="{'fa-check': !autocompile_enabled}")
+						| &nbsp;Off
 				li.dropdown-header #{translate("compile_mode")}
 				li
 					a(href, ng-click="draft = false")
--- a/app/views/project/editor/pdf.pug
+++ b/app/views/project/editor/pdf.pug
@@ -1,4 +1,5 @@
 div.full-size.pdf(ng-controller="PdfController")
+	.FIXME(ng-controller="AutoCompileController")
 	.toolbar.toolbar-tall
 		.btn-group(
 			dropdown,
--- a/app/views/project/editor/pdf.pug
+++ b/app/views/project/editor/pdf.pug
@@ -1,5 +1,4 @@
 div.full-size.pdf(ng-controller="PdfController")
-	.FIXME(ng-controller="AutoCompileController")
 	.toolbar.toolbar-tall
 		.btn-group(
 			dropdown,
--- a/public/coffee/ide/editor/directives/aceEditor.coffee
+++ b/public/coffee/ide/editor/directives/aceEditor.coffee
@@ -376,6 +376,10 @@ define [
 
 					$rootScope.hasLintingError = false
 					session.on('changeAnnotation', () ->
+						# Both linter errors and compile logs are set as error annotations,
+						# however when the user types something, the compile logs are
+						# replaced with linter errors. When we check for lint errors before
+						# autocompile we are guaranteed to get linter errors
 						hasErrors = session
 							.getAnnotations()
 							.filter((annotation) -> annotation.type == 'error')
--- a/public/coffee/ide/editor/directives/aceEditor.coffee
+++ b/public/coffee/ide/editor/directives/aceEditor.coffee
@@ -378,7 +378,7 @@ define [
 					session.on('changeAnnotation', () ->
 						hasErrors = session
 							.getAnnotations()
-							.filter((annotation) -> annotation.type == 'error')
+							.filter((annotation) -> annotation.fromLinting && annotation.type == 'error')
 							.length > 0
 
 						if ($rootScope.hasLintingError != hasErrors)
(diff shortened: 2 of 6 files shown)


public/coffee/ide/editor/directives/aceEditor.coffee
```diff
@@ -374,6 +374,18 @@ define [
...
 
+					hasLintingError = null
+					session.on('changeAnnotation', () ->
+						hasErrors = session
+							.getAnnotations()
+							.filter((annotation) -> annotation.type == 'error')
+							.length > 0
+
+						if (hasLintingError != hasErrors)
+							ide.$scope.$broadcast('ide:lintingError', hasErrors)
```
jpallen: We already have a reference to `$rootScope`, so we shouldn't need `ide` here.

app/views/project/editor/pdf.pug
```diff
@@ -1,4 +1,5 @@
 div.full-size.pdf(ng-controller="PdfController")
+	.FIXME(ng-controller="AutoCompileController")
```
jpallen: I guess we might need some UI for this at some point that will go here? But if not, this logic probably shouldn't live in a controller

public/coffee/ide/editor/directives/aceEditor.coffee
jpallen: In fact, it should be possible to just trigger an event on our own `scope`, not even `$rootScope`

jpallen: See https://github.com/sharelatex/web-sharelatex/blob/2ef8d29a907259f5080473d93c881c80f0b5bec5/public/coffee/ide/editor/directives/aceEditor.coffee#L304 for an example, which also namespaces the event under the editor name, which is important because there are multiple Ace editors in the app (one is also used for the history view)

public/coffee/ide/editor/directives/aceEditor.coffee
```diff
@@ -374,6 +374,18 @@ define [
 					if scope.eventsBridge?
 						session.on "changeScrollTop", onScroll
 
+					hasLintingError = null
+					session.on('changeAnnotation', () ->
+						hasErrors = session
```
jpallen: Will this not all pick up errors from the compile logs (which are injected into the editor as annotations as well)?

public/coffee/ide/editor/directives/aceEditor.coffee
40thieves: Ah you're right. For some reason I thought pdf and editor were in different scopes. Also changed the event name to namespace under the editor.

app/views/project/editor/pdf.pug
40thieves: A controller is what Brian suggested, but yes I think it's simpler to just pull it into the `PdfController`

public/coffee/ide/editor/directives/aceEditor.coffee
40thieves: As discussed, I've added an extra flag to linting output (https://github.com/sharelatex/ace/pull/4), which we then filter on when listening for changes

40thieves:
⚠️  Cleaned up history & force pushed

public/coffee/ide/pdf/controllers/PdfController.coffee
```diff
@@ -73,6 +73,11 @@ define [
 			$scope.pdf.view = 'errors'
 			$scope.pdf.renderingError = true
 
+		ide.$scope.$on "ide:opAcknowledged", _.debounce(() ->
+			if (!ide.$scope.hasLintingError)
+				$scope.recompile()
+		, 3000)
```
40thieves: 3 seconds felt about right, but happy to change this if needed

public/coffee/ide/pdf/controllers/PdfController.coffee
```diff
@@ -73,6 +74,27 @@ define [
 			$scope.pdf.view = 'errors'
 			$scope.pdf.renderingError = true
 
+		triggerAutoCompile = () ->
+			if (!ide.$scope.hasLintingError)
+				$scope.recompile()
+
+		autoCompileListener = null
+		toggleAutoCompile = (enabling) ->
```
40thieves: I'm not sure if I made this overly complicated but I wanted to try tearing down the listener if auto-compile is disabled. This prevents unnecessary calls, but might be pre-optimisation.

Happy to remove if you disagree

public/coffee/ide/pdf/controllers/PdfController.coffee
```diff
@@ -5,8 +5,9 @@ define [
 	"libs/bib-log-parser"
 	"services/log-hints-feedback"
 ], (App, Ace, HumanReadableLogs, BibLogParser) ->
-	App.controller "PdfController", ($scope, $http, ide, $modal, synctex, event_tracking, logHintsFeedback, localStorage) ->
+	AUTO_COMPILE_TIMEOUT = 3000
 
+	App.controller "PdfController", ($scope, $http, ide, $modal, synctex, event_tracking, logHintsFeedback, localStorage) ->
 		# enable per-user containers by default
 		perUserCompile = true
 		autoCompile = true
```
40thieves: I'm a bit worried about overloading the term "auto compile", as we already use it to mean "auto compile on initial page load" (at least AFAICT). I was struggling with alternative naming though

briangough:
To simplify, I don't think we need the fromLinting flag as the editor behaviour will give the same result.  What already happens is as soon as you start typing any existing annotations from the log file are replaced by the linter ones, so the log errors can be ignored for this purpose, because we won't trigger an autocompile unless the user edits something.

I agree about the name autocompile, but perhaps we have to see the initial autocompile and the editing driven ones as part of the same system.

briangough:
Can we make the recompile behaviour the same as Overleaf, which - as I understand it - is compile after  5 seconds idle time or 20 seconds otherwise.

public/coffee/ide/pdf/controllers/PdfController.coffee
briangough: Yup it's good to clean things up.

jdleesmiller:
> is compile after 5 seconds idle time or 20 seconds otherwise.

Correct. When implementing that, we should also try not to replicate the bug we had recently with non-monotonic clocks. Internal ref: https://github.com/overleaf/write_latex/issues/3573

briangough:
Re clocks, ouch!

Ali, to add analytics, you only need to do one thing (1) below, everything else is automatic.  The two places where the events happen are:-

1.  In the recompile method - you just need to pass an option in (e.g. {autoCompile:true}) and it will get sent to the analytics system

https://github.com/sharelatex/web-sharelatex/blob/01e22fd982cfcff8e8609054c3bf76bd2a4c5c3e/public/coffee/ide/pdf/controllers/PdfController.coffee#L370

2. Settings in left menu  - this is already handled automatically by the controller (https://github.com/sharelatex/web-sharelatex/blob/master/public/coffee/ide/settings/controllers/SettingsController.coffee) which has automatically sends analytics events when saving any setting 

https://github.com/sharelatex/web-sharelatex/blob/01e22fd982cfcff8e8609054c3bf76bd2a4c5c3e/public/coffee/ide/settings/services/settings.coffee#L11

https://github.com/sharelatex/web-sharelatex/blob/01e22fd982cfcff8e8609054c3bf76bd2a4c5c3e/public/coffee/ide/settings/services/settings.coffee#L22

If you need to know more about what the analytics system can do @paulojreis can fill you in.

40thieves:
> compile after 5 seconds idle time or 20 seconds otherwise.

So I just had a chat with @jdleesmiller and we worked out that OL has a slightly more nuanced debounce. Partly this is due to the fact that we debounce the save operation as well as compiles, but also it's smarter about kicking off compiles quicker if you haven't edited for a while.

The key is this function: https://github.com/overleaf/write_latex/blob/56e35ab12ba63610a9fa67988c665bde70475152/main/app/assets/javascripts/wl-listener.js#L149

The user has to be idle for 5 seconds before we start a recompile. This acts like a rate limiter. However, if we haven't recompiled for 5 or more seconds, then the recompile will start immediately. This way we're quicker to recompile if the user is just making small changes. But if they're typing a lot, then we rate limit to a minimum of 5 seconds. Hopefully that makes sense, it's a bit difficult to explain.

At this point I think the best plan is to merge what we have here (although I'll change to 5 seconds), then loop back around on implementing the above. What we have works, the above is really just a small UX improvement.

> I agree about the name autocompile, but perhaps we have to see the initial autocompile and the editing driven ones as part of the same system.

Do you think we should include the initial autocompile in the setting too then?  Although I foresee this having problems with existing users.

> To simplify, I don't think we need the fromLinting flag as the editor behaviour

That makes sense to me. @jpallen originally suggested it, so I just want to check with him too.

40thieves:
Only just seen the extra comment...

> non-monotonic clocks

I'll make sure to check for that too.

> Ali, to add analytics

Thanks, that looks like a useful starting point

jpallen:
No problem on my end with removing the `fromLinter` flags if not needed. I might have sent you on a wild goose chase there, sorry. I'd only assumed there would be problems because the linter isn't the only thing that adds annotations, but it may be fine in practice - if so can you add a comment note explaining why the PDF annotations aren't a problem.

briangough:
> Do you think we should include the initial autocompile in the setting too then? Although I foresee this having problems with existing users.

No need to combine them, just meant that conceptually they're similar so I don't think it's a big problem for the names to be the same.

40thieves:
Pushed some changes: increased timeout to 5 seconds, inline with OL and removed the unnecessary linting check.

I also added tracking for when the auto recompile is triggered. Unfortunately we're already using a `isAutoCompile` option: https://github.com/sharelatex/web-sharelatex/blob/e45a2da8fae3705e2a418e6231f2e8762dcd23a1/public/coffee/ide/pdf/controllers/PdfController.coffee#L70 for the initial autocompile. So I decided to name it `isBackgroundAutoCompile`. Suggestions on something better welcomed.

> Settings in left menu - this is already handled automatically by the controller 

The thing is that we're not using the left menu for this setting - it's in the dropdown on the compile button. So I've added in the tracking now.

public/coffee/ide/editor/directives/aceEditor.coffee
```diff
@@ -374,6 +374,21 @@ define [
...
 
+					$rootScope.hasLintingError = false
+					session.on('changeAnnotation', () ->
+						# Both linter errors and compile logs are set as error annotations,
+						# however when the user types something, the compile logs are
+						# replaced with linter errors. When we check for lint errors before
+						# autocompile we are guaranteed to get linter errors
+						hasErrors = session
+							.getAnnotations()
+							.filter((annotation) -> annotation.type == 'error')
```
briangough: typing a command with an unclosed bracket like `\foo{...` comes up as a warning `unclosed group {` while the user is typing it.

I think it's better if we also don't run autocompile for `type == 'warning'` as well as `'error'`  as this will not give a good result when the group hasn't been closed.   

We can filter out `type == 'info'` which only happens for things like content after \end{document}

40thieves: Good point. Pushed a fix

40thieves:
Merged #598 into here

jpallen:
Closing and tracking at https://github.com/sharelatex/web-sharelatex-internal/pull/4

PR #593, opened 2017-09-05
avoid error when passing as limit in mongo query

briangough:
convert request parameter from string to number


briangough:
noticed this was triggering an error in sentry -- 
https://sentry.io/sharelatex-1/sl-web-server-prod/issues/320033548/

PR #596, opened 2017-09-07
suppress incremental compile after errors

briangough:
to allow recovery from problems in the incremental updates, do a full compile after any server error from the clsi or error getting incremental files from the docupdater.

Currently, if there is an error from either of those we keep trying the incremental request - with no way for the user to recover.

### DIFF
--- a/app/coffee/Features/Compile/ClsiManager.coffee
+++ b/app/coffee/Features/Compile/ClsiManager.coffee
@@ -129,10 +129,15 @@ module.exports = ClsiManager =
 				timer = new Metrics.Timer("editor.compile-getdocs-redis")
 				ClsiManager.getContentFromDocUpdaterIfMatch project_id, project, options, (error, projectStateHash, docUpdaterDocs) ->
 					timer.done()
-					return callback(error) if error?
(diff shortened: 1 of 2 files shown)


app/coffee/Features/Compile/ClsiManager.coffee
```diff
@@ -129,10 +129,15 @@ module.exports = ClsiManager =
...
+					if error?
+						logger.error err: error, project_id: project_id, "error checking project state"
+						# note: we don't bail out when there's an error getting
+						# incremental files from the docupdater, we just fall back
+						# to a normal compile below
+					else
+						logger.log project_id: project_id, projectStateHash: projectStateHash, docs: docUpdaterDocs?, "checked project state"
 					# see if we can send an incremental update to the CLSI
-					if docUpdaterDocs? and options.syncType isnt "full"
+					if docUpdaterDocs? and options.syncType isnt "full" and not error?
```
jpallen: Nitpick: can you add some brackets: `if docUpdaterDocs? and (options.syncType isnt "full") and not error?`. Otherwise it's a bit of a garden path sentence for me.

PR #597, opened 2017-09-07
Add security headers

timothee-alby:
### Description

Adds a bunch of security headers to better protect the app from some attacks.
This uses the [Helmet package](https://helmetjs.github.io/docs/) see the documentation for more details about each headers.

##### Helmet defaults, disabled:

- `X-DNS-Prefetch-Control`
  I don't see any good reason to disable prefetching

##### Helmet defaults, already in use on sharelatex.com:

- `X-Frame-Options: SAMEORIGIN`
- `Strict-Transport-Security: max-age=15552000; includeSubDomains`
- `X-Content-Type-Options: nosniff`

No impact expected.

##### Helmet defaults, added:

- disable `X-Powered-By`
- `X-Download-Options: noopen`
  already used on OL
- `X-XSS-Protection: 1; mode=block`
  already used on OL

Potential Impact: low, should only block features we don't use.

##### Helmet non-default, added:

- `Surrogate-Control: no-store` 
- `Cache-Control: no-store, no-cache, must-revalidate, proxy-revalidate` (already used on OL)
- `Pragma: no-cache` 
- `Expires: 0` (already used on OL, albeit differently)

Those 4 headers disable page caching on client's end when it may contain sensitive information. This is mainly to prevent unauthorized access by an attacker using e.g. the browser's back button.
Potential impact: performances as it will generate (relatively low) extra load on the servers as some requests won't be cached anymore 

- `Referrer-Policy: origin-when-cross-origin` (already used on OL)

Don't disclose query string to third-party
Potential impact: information shared with partners, analytics

#### Related Issues / PRs

https://github.com/overleaf/sharelatex/issues/8

#### Potential Impact

See description.

#### Testing Checklist

- [x] acceptance tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "express": "4.13.0",
     "express-session": "^1.14.2",
     "heapdump": "^0.3.7",
+    "helmet": "^3.8.1",
     "http-proxy": "^1.8.1",
     "ioredis": "^2.4.0",
     "jade": "~1.3.1",
--- a/app/coffee/infrastructure/Server.coffee
+++ b/app/coffee/infrastructure/Server.coffee
@@ -6,6 +6,7 @@ metrics = require('metrics-sharelatex')
 crawlerLogger = require('./CrawlerLogger')
 expressLocals = require('./ExpressLocals')
 Router = require('../router')
+helmet = require "helmet"
 metrics.inc("startup")
 UserSessionsRedis = require('../Features/User/UserSessionsRedis')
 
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "express": "4.13.0",
     "express-session": "^1.14.2",
     "heapdump": "^0.3.7",
+    "helmet": "^3.8.1",
```
ShaneKilkelly: Just a note, you'll need to run `npm shrinkwrap` to update the `npm-shrinkwrap.json` file, and commit those changes too.

NPM uses the shrinkwrap file if it's present, and doesn't look at the package.json file. I suspect in the current state it wouldn't actually install helmet.

ShaneKilkelly: Best way to test this is to `rm -rf node_modules && npm install`, then try to run the app.

timothee-alby: Thanks, I missed that!

timothee-alby:
Updated the `npm-shrinkwrap.json` file.

⚠️ rebased on top of master and force-pushed

jdleesmiller:
Chiming in uninvited... but usually when I install a new package I do it with `npm install --save foo`, and I'm pretty sure that `npm install` does update the shrinkwrap, at least in recent versions of npm. Is that also how you've been doing it, @ShaneKilkelly / @timothee-alby ?

timothee-alby:
Not sure how I did it initially, but the shrinkwrap wasn't updated.. This time Shane suggested I remove the shrinkwrap file first, then re-generate it

ShaneKilkelly:
It's possible that I'm remembering an earlier (and much worse) version of npm where the shrinkwrap was not updated when `--save`ing a new package. Probably worth trying it that way (the way John suggested) too.

timothee-alby:
It doesn't look like there's much difference; it still produce a huge, messy diff for the shrinkwrap file.

Anyways, updated

PR #603, opened 2017-09-12
Pin chai version

ShaneKilkelly:


jpallen:
Has this caught you out with something? Seems like a good idea, but what's the background?

ShaneKilkelly:
Yeah, it got bumped on master and now the unit tests are failing, because of the good ol' error-equality problem. This change gets it back to where it was yesterday.

We can later then think about actually upgrading to the latest chai.

PR #604, opened 2017-09-12
fix root doc in incremental compile

briangough:
This is just the proof of concept - not tested yet!


briangough:
@jpallen thought you might want to see this - haven't actually run/tested it.   Main point is, we already have all the information in web to identify the root doc path without additional calls to mongo, so no need to save it on the clsi or anything complicated.

jpallen:
LGTM, seems like a clean approach to go ahead and test etc

briangough:
Updated with tested version

PR #605, opened 2017-09-13
Remove extra security headers

timothee-alby:
Don't add `Strict-Transport-Security`, `X-Content-Type-Options` and `X-Frame-Options` headers.


jdleesmiller:
Could you add some notes on this when ready?

Interested to know whether they caused a problem.

jpallen:
Did some of these causes issues with web-api (it's accessed over HTTP)? Has the impact on web-api been considered?

briangough:
The headers affect the browser, so shouldn't be a problem for internal requests.

timothee-alby:
Follow-up here: https://github.com/overleaf/sharelatex/issues/173

I realised the issue on staging, so no impact on production

PR #662, opened 2019-04-29
[docker] add a whitelist for the image files

das7pad:
Dropping the `.git` directory and some dev files lowers the image size of the final app image by about 63MB.


das7pad:
The allow list is too strict for the internal repo.

PR #663, opened 2019-04-29
[docker] ci: cache the google-chrome install stage

das7pad:
This PR adds support for caching of the google chrome installation stage in the CI image.

The docker instruction `COPY` has not support for variables in the `--from` argument [1], using a named stage in front of the final image solves this.

---
[1] See upstream issue https://github.com/moby/moby/issues/34482


das7pad:
Reusing the app layer and pulling chrome is actually faster with a decent internet connection.

PR #664, opened 2019-04-29
[docker] drop a useless chmod and chown in the build stage

das7pad:
In dropping the chmod/chown instruction we can lower the image size by about 73MB.

File content and file attributes are coupled together in docker. Changing any attribute of a file - e.g. the owner or permissions - results in a new entry for the file in another layer of image.

The CI passed here, so I assume it was not needed anyways. Otherwise we should refactor the install script into two parts. One for the build layer with the given chmod plus co and one for the final layer that creates the user data directories.


das7pad:
Super seeded by 1346a3124bd8dc2c136f606ae19a1965bf47c4aa

PR #666, opened 2019-04-29
[make] simplify the module compile targets

das7pad:
This PR introduces a simple fan out of build targets to the modules.

The build target `modules/<MODULE>/<TARGET>` translates into the build target  `<TARGET>` in the given `<MODULE>`.

For example `make modules/launchpad/compile` runs the `compile` target in the `launchpad` module.

The `compile` and `compile_full` targets are already migrated, others may need some additional tweeks. See following PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,30 +148,11 @@ compile_full:
 compile_css_full:
 	$(MAKE) css_full
 
-compile_modules: $(MODULE_MAKEFILES)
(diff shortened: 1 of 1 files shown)


das7pad:
superseded by e6b5e37ae3be044333360c0d89393efd255d19a3 8fe0e9bfff66989b443b6139693f4b205573f6d2

PR #667, opened 2019-04-29
[make] simplify the acceptance tests run for modules

das7pad:
This is a followup for #666. The actual diff is https://github.com/sharelatex/web-sharelatex/compare/be0ad70cdc11fa3e505d5cf6e3de3358a935128d...a6261e6aeea2aeb6850c58b3f0841ee9a24086a7.

This PRs adds a `clean_test_acceptance` target into the module Makefile.

Additionally it uses the new fan out method of #666 for the test invocation and cleanup.

Modules without acceptance tests get empty build targets for the test run and cleanup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,30 +148,11 @@ compile_full:
 compile_css_full:
 	$(MAKE) css_full
 
-compile_modules: $(MODULE_MAKEFILES)
-	@set -e; \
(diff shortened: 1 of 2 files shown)


das7pad:
Landed in https://github.com/overleaf/web/commit/b0c60951207ad258172849ffbf6055a426194869

PR #669, opened 2019-04-29
[make] copy the module Makefile only once per module

das7pad:
As per the title.


das7pad:
Landed in https://github.com/overleaf/web/commit/b0c60951207ad258172849ffbf6055a426194869

PR #670, opened 2019-05-02
[logging] do not overwrite the logger name

das7pad:
Similar to https://github.com/sharelatex/filestore-sharelatex/pull/50.

The user provided document or filename overwrites the logger name in some cases.

This PR moves the names into dedicated fields.


das7pad:
You can scan for bad logger field names in your other projects with this regex: `logger.*\bname:`.

das7pad:
Landed https://github.com/overleaf/web/commit/99eb3f9025019632008650b55700f9c37d9adab7

PR #675, opened 2019-06-12
[misc] readme: fix a typo in the name of the parent repository

das7pad:
As per the title.


das7pad:
Landed in https://github.com/overleaf/web/commit/c2825fa10f2de93020982dfd1716e7916f8189d7

jdleesmiller:
Yep, thanks. We tried to preserve your authorship on the commit, but it didn't work, sorry. We have a new approach that we're trying with one of your later PRs.

PR #690, opened 2019-10-25
[model] User: add missing unique constraint on the email field

das7pad:
One of the Model acceptance tests [1] hints on a unique constraint for the email field of the User model, but it is missing in the model spec. This PR adds said constraint.

The mongodb connection for the user model has the `autoIndex` flag disabled [2], so this should not have an impact on the regular usage of this model.

---
[1] https://github.com/overleaf/web/blob/c6bf5eb8e161d2a1017c06c38ac5837974872d70/test/acceptance/src/ModelTests.js#L18-L22

[2] https://github.com/overleaf/web/blob/c6bf5eb8e161d2a1017c06c38ac5837974872d70/app/src/models/User.js#L132

### DIFF
--- a/app/src/models/User.js
+++ b/app/src/models/User.js
@@ -16,7 +16,12 @@ const { ObjectId } = Schema
 const MAX_EMAIL_LENGTH = 254
 
 const UserSchema = new Schema({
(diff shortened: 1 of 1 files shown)


das7pad:
Migrations are public now.

PR #693, opened 2019-10-26
[misc] do not force a scrollbar on the project list view

das7pad:
The actual project list (wrapper) does not overflow.

Git blame on the line suggests that optional SSO notifications may
 cause an overflow.

Let's hide the scrollbar by default and only show it when necessary.

REF: ebb6cc5adeab5135fb65a5401535e1c5fe320c1e
cc @lawshe

### DIFF
--- a/public/stylesheets/app/project-list.less
+++ b/public/stylesheets/app/project-list.less
@@ -90,7 +90,7 @@
   padding-bottom: @content-margin-vertical;
   height: 100%;
   margin-left: -(@grid-gutter-width / 2);
-  overflow-y: scroll;
+  overflow-y: auto;
 }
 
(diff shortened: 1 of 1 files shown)


das7pad:
Landed in https://github.com/overleaf/web/commit/038bef4de4b5ffd020dc9e1757e08e9010272b83

lawshe:
Thank you @das7pad! It looks like the commit didn't include you as an author when we merged it in. I am very sorry and we will look into getting that fixed for the future.

das7pad:
No worries. The copy bot is just cherry picking a sqash. This is also a problem for the decaffeinate merge for example, which breaks the git history for the backend files.

PR #701, opened 2019-11-15
Added 'csv' to list of editable text file extensions

samusangio:
This allows files with extension '.cvs' to be editable in the overleaf editor


das7pad:
Thank you for your contribution. In case you are still interested in bringing this in, please open a new PR at our public monorepo https://github.com/overleaf/overleaf.
The list of editable text file extensions has been moved here: https://github.com/overleaf/overleaf/blob/94117ea836ccf1d1a8d46fa79febc604fd8444a2/services/web/config/settings.defaults.js#L32-L64

PR #707, opened 2019-12-13
Feature relative image paths

das7pad:
### Description
Images are currently hard linked to the root of its serving origin. Using relative paths we can serve multiple versions of the assets from the same cdn domain - or use a multi purpose cdn as in `https://cdn.example.com/overleaf/`. Additionally some IDEs will allow an easy navigation from the css rule to the actual image.

Webpack will error out in case it can not find a linked image.

### Review
- https://github.com/overleaf/web/commit/8fbd8935f028dfc1201f3c0fcf28e240edfe8878 should be trivial.

- https://github.com/overleaf/web/commit/eff51a0cb5d3f411bbfb044de886a643ed82adc9 is slightly hacky but I could not find a simpler option that does A) not emit a new file and B) preserves to uri of the resources. Just using the `file-loader` would create new files like `fd2a63eeee604ff679003c8c28c41386.svg` in the public/ directory and like these from css instead. Using the ignore plugin does not work as it would complain for the missing resources.

#### Manual Testing Performed

- [ ]
- [ ]

#### Deployment Checklist

- [ ]

### DIFF
--- a/frontend/stylesheets/app/editor.less
+++ b/frontend/stylesheets/app/editor.less
@@ -232,9 +232,9 @@ Ace
   }
   .spelling-highlight {
     position: absolute;
-    background-image: url(/img/spellcheck-underline.png);
+    background-image: url(../../../public/img/spellcheck-underline.png);
     @media (-webkit-min-device-pixel-ratio: 2), (min-resolution: 192dpi) {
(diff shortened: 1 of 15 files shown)


das7pad:
Needs to be reworked for latest changes and rebased into monorepo.

PR #709, opened 2019-12-13
[misc] fileOperationI18nNames: drop blocking DOM access

das7pad:
### Description
`.innerText` is a blocking dom operation and requires e.g. font files to be loaded already.

The four strings are better put into a single json blob and parsed back in the front-end js.

#### Manual Testing Performed

- [ ]
- [ ]

#### Deployment Checklist

- [ ]

### DIFF
--- a/app/views/project/editor/history/fileTreeV2.pug
+++ b/app/views/project/editor/history/fileTreeV2.pug
@@ -56,7 +56,6 @@ script(type="text/ng-template", id="historyFileEntityTpl")
 				file-entity="childEntity"
 			)
 
-script(type="template", id="file_action_edited_str") #{ translate('file_action_edited') }
-script(type="template", id="file_action_renamed_str") #{ translate('file_action_renamed') }
-script(type="template", id="file_action_created_str") #{ translate('file_action_created') }
-script(type="template", id="file_action_deleted_str") #{ translate('file_action_deleted') }
+- var fileActionI18n = ['edited', 'renamed', 'created', 'deleted'].reduce((acc, i) => {acc[i] = translate('file_action_' + i); return acc}, {})
+script(type="text/javascript").
+	window.fileActionI18n = JSON.parse('!{StringHelper.stringifyJsonForScript(fileActionI18n)}')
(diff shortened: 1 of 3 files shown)


app/views/project/editor/history/fileTreeV2.pug
```diff
@@ -56,7 +56,6 @@ script(type="text/ng-template", id="historyFileEntityTpl")
 				file-entity="childEntity"
 			)
 
-script(type="template", id="file_action_edited_str") #{ translate('file_action_edited') }
-script(type="template", id="file_action_renamed_str") #{ translate('file_action_renamed') }
-script(type="template", id="file_action_created_str") #{ translate('file_action_created') }
-script(type="template", id="file_action_deleted_str") #{ translate('file_action_deleted') }
+- var fileActionI18n = ['edited', 'renamed', 'created', 'deleted'].reduce((acc, i) => {acc[i] = translate('file_action_' + i); return acc}, {})
```
40thieves: I think a `map` would be more appropriate here? Something like:

```suggestion
- var fileActionI18n = ['edited', 'renamed', 'created', 'deleted'].map((action) => translate('file_action_' + action))
```

I also made the variable bit more descriptive.

das7pad: I agree with you, the variable names are not very descriptive.

Your suggestion yields an array with the translated values, but we need a `Map` with entries `action -> translate('file_action_' + action)`.
```
> ['edited', 'renamed', 'created', 'deleted'].reduce((acc, i) => {acc[i] = 'file_action_' + i; return acc}, {})
{ edited: 'file_action_edited',
  renamed: 'file_action_renamed',
  created: 'file_action_created',
  deleted: 'file_action_deleted' }
>
> ['edited', 'renamed', 'created', 'deleted'].map((action) => 'file_action_' + action)
[ 'file_action_edited',
  'file_action_renamed',
  'file_action_created',
  'file_action_deleted' ]
```

40thieves: Ah right, sorry I didn't spot that. Happy to keep the `reduce` then.

das7pad:
Landed in https://github.com/overleaf/web/commit/326d8d46d2a1e92d3ea0cea2ece8d9bdcf682fde

PR #723, opened 2020-08-03
add lastUpdated field to route /user/projects

monperrus:
### Description

Add lastUpdated field to route /user/projects. This is required to make a resource aware incremental backup. Will reduce the load on overleaf's servers.

#### Manual Testing Performed

- [ ]
- [ ]

#### Deployment Checklist

- [ ]

### DIFF
--- a/app/src/Features/Project/ProjectController.js
+++ b/app/src/Features/Project/ProjectController.js
@@ -342,7 +342,7 @@ const ProjectController = {
         // _buildProjectList already converts archived/trashed to booleans so isArchivedOrTrashed should not be used here
         projects = ProjectController._buildProjectList(projects, userId)
           .filter(p => !(p.archived || p.trashed))
-          .map(p => ({ _id: p.id, name: p.name, accessLevel: p.accessLevel }))
+          .map(p => ({ _id: p.id, name: p.name, accessLevel: p.accessLevel, lastUpdated: p.lastUpdated }))
 
         res.json({ projects })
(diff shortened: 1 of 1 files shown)


monperrus:
ping @lawshe

gh2k:
Thank you for taking the time to submit this.

We generally prefer to keep data out of the rendered pages unless it is going to be used by something in the UI, and it looks like the reason why you're looking for this is for some kind of external script, which isn't something we officially support right now, so I'm going to close this for now.

I've made a note of your use case and this might be something we can look into again in the future.

monperrus:
Thanks a lot @gh2k for your answer

I don't find this information anywhere in the doc. 

Can we get this information on a project specific end-point? 

Something like

```
/project/<projectid>/XXXX

{
  "id" :  "<projectid>",
  "lastUpdated" : "2020-12-12"
  ...
}
```

WDYT?

Thanks! --Martin

gh2k:
Thanks for this idea. We do have conversations internally about what we might provide as part of an API, and this is definitely a use case for that.

Unfortunately, we're still not going to add something to the main app to support an external use case without considering it as part of a larger piece of work on making a documented API. Otherwise, what we end up with is a bunch of undocumented things that we then have to maintain for external users.

It *is* a good idea, but not something that we are going to implement for now. Sorry about that.

monperrus:
OK thanks Simon. --Martin

PR #724, opened 2020-08-16
Avoid JSON indirect parsing

zauguin:
### Description
Replace `JSON.parse('!{StringHelper.stringifyJsonForScript(ExposedSettings)}')` by just `!{StringHelper.stringifyJsonForScript(ExposedSettings)}` because the old construction is useless (JSON is valid JS anyway) and doesn't escape the object properly (`'` in strings lead to errors)

#### Manual Testing Performed

- [x] Deployed on personal Overleaf Community Edition install where it fixed a bug related to a single quote `'` in APP_NAME
- [ ]

#### Deployment Checklist

- [ ]

### DIFF
--- a/app/views/layout.pug
+++ b/app/views/layout.pug
@@ -75,7 +75,7 @@ html(
 			};
 			window.ab = {};
 			window.user_id = '#{getLoggedInUserId()}';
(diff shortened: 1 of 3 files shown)


das7pad:
Thank you for your contribution. We have migrated to a different mechanism for passing JSON data to the frontend. It can handle any kinds of quotes (in settings) now.

PR #748, opened 2021-02-20
Resolve #659: Relative line numbers for vim

### LINKED ISSUE #659: [Feature request] Relative line number
As a vim user, relative line number is so important that it kind of defeats the purpose of using vim without it. I believe ace supports relative line number and hope sharelatex can integrate it.

Edit: It is only available on v1.4.x, can only be done with #652

lhhong:
### Description
This is a quick and simple solution without adding fine tuned controls for line number display. This forces vim mode to use relative line numbers for simplicity because honestly, I don't think any vim users would ever want absolute line numbers...

#### Related Issues / PRs
#659 

#### Manual Testing Performed

- [ ]
- [ ]

#### Deployment Checklist

- [ ]

### DIFF
--- a/frontend/js/ide/editor/directives/aceEditor.js
+++ b/frontend/js/ide/editor/directives/aceEditor.js
@@ -455,10 +455,16 @@ App.directive('aceEditor', function(
       scope.$watch('showPrintMargin', value => editor.setShowPrintMargin(value))
 
       scope.$watch('keybindings', function(value) {
-        if (['vim', 'emacs'].includes(value)) {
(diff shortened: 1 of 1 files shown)


das7pad:
Thank you for your contribution. I've forwards this to our product team for consideration. We are in the process of migrating to a monorepo (https://github.com/overleaf/overleaf) and we need to close this PR for the time being. https://github.com/overleaf/web/issues/659 is on our radar and we might address it in the future (I cannot provide any timeline tho).

PR #768, opened 2021-05-20
Fix file naming for PDF downloads

cmprmsd:
Our documents contain `@` as well as `-` symbols. We do not consider them harmful and projects can have the symbols in their name.
As copies of projects have (1) in their name, it would also make sense to include them into the regular expression. `...replace(new RegExp('[^\\(\\)\\@\\-\\w]', 'g'), '_')`

#### Manual Testing Performed

- [ ]
- [ ]

#### Deployment Checklist

- [ ]

### DIFF
--- a/app/src/Features/Compile/CompileController.js
+++ b/app/src/Features/Compile/CompileController.js
@@ -264,8 +264,8 @@ module.exports = CompileController = {
   },
 
   _getSafeProjectName(project) {
-    const wordRegExp = /\W/g
-    const safeProjectName = project.name.replace(wordRegExp, '_')
+    // Only allow alphanumeric and ( ) @ - characters in filenames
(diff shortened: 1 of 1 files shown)


das7pad:
Thank you for your contribution. In case you are still interested in bringing this in, please open a new PR at our public monorepo https://github.com/overleaf/overleaf. The code of the web service is located in `services/web/`.
Could you also add some tests for the changed regex? (You could amend some more special characters to the project name and update the assertion on the `res.setContentDisposition` arguments [here](https://github.com/overleaf/overleaf/blob/94117ea836ccf1d1a8d46fa79febc604fd8444a2/services/web/test/unit/src/Compile/CompileControllerTests.js#L268-L297).)

cmprmsd:
I don't know how to write tests :/ 
Not a coder. I just tried to contribute to the project. 😊

PR #769, opened 2021-05-23
fix bugs & add features

Makiras:
read commits messages


Makiras:
So sorrry for wrong target selected

PR #773, opened 2021-06-30
Fix translation issues in zh-CN

jnhu76:
### Description

This word is incorrectly translated, part of the word is repeatedly translated.

`nearly_activated`: `还有一步您的 __appName__ 账户就会被激活了！`.


jdleesmiller:
Thanks for reporting this. We use a translation service to manage the locales, so editing the files directly doesn't work, but I have updated it upstream, and it landed in https://github.com/overleaf/web/commit/c404d86d92a8e2084d736bef592f74d46b67234d . Thanks again!

